PR #82, opened 2017-11-02
Screen stack base

AttwellBrian:
Uber's implementation of `ScreenStack` is quite complicated. It includes analytics and various forms of (infrequently used) animation customization. I don't want to open source the full implementation because (1) that would be difficult (2) a lot of the functionality doesn't make sense as a general utility.

This is not a mandatory part of the RIB framework. But it is an interesting glimpse into the way RIBs can be used. So I figured I'd include the interface in the initial release plus a short readme


rendecano:
Hi,

Any updates on the tutorial for ScreenStack? :)

Thanks!

bangarharshit:
@rendecano a very simple screenstack will be a just stack of view provider - https://gist.github.com/bangarharshit/a7b4614bd5d2fc77171fca8cdbb11f8b. You still need to add some state restoration (for the case when an old view pops up from stack) - http://trickyandroid.com/saving-android-view-state-correctly/

rendecano:
Thanks @bangarharshit!

PR #91, opened 2017-11-04
Fix travis for iOS

artman:


.travis.yml
```diff
@@ -38,15 +36,6 @@ matrix:
         - WORKSPACE=ios/RIBs.xcworkspace
```
sbarow: Looks like the workspace dir is incorrect. It should just be `- WORKSPACE=RIBs.xcworkspace` as we do a `cd ios`

PR #95, opened 2017-11-04
Updating RIB code generation icon

gergelyorosz:
Updating the icon to match the RIB icon, as well as the binary jar to reflect this change. Note that the 16x16px icon doesn't look that great, just scaling it down. If someone could come up with a better one, that would be great.

Alternatively, we can keep the old "RIB" icon until we get a better low-res icon as well. The iOS toolbar has a lot higher resolution, hence the default RIBs icon looks better there.

If we decide to replace the icon, I'll re-generate affected screenshots in the tuturials.

screen shot 2017-11-04 at 2 56 04 pm


AttwellBrian:
you may want to rebuild the intellij plugin and replace the binary in the repo if you want people to be able to use this.

gergelyorosz:
Thanks, Brian. I've rebuilt the plugin. Updating the description to make this clear.

Issue #100, opened 2017-11-05
intellij module build performance


AttwellBrian:
The build depends on `com.uber.intellij` dependencies. Obviously these are not open source dependencies.

Didn't notice until just now. Will need to find a replacement for these.

FYI @jbarr21 @kageiit

AttwellBrian:
This is now fixed. However the build will be really slow because hundreds of MBs are downloaded from https://www.jetbrains.com/intellij-repository/snapshots. 

Should look into dropping the exact jars that we need.

PR #101, opened 2017-11-05
Re-enable intellij module

AttwellBrian:
Previously I needed to disable the intellij build in order to make travis green. This was necessary because the `com.uber.intellij:xxx` dependencies didn't exist in the oss world.

Realized that I could avoid using the `com.uber.intellij:xxx` dependencies if I simply applyied the intellij plugin. Granted this makes the regular build & gradle sync a lot slower (takes a few minutes to download from https://www.jetbrains.com/intellij-repository/snapshots).

For https://github.com/uber/RIBs/issues/100


kageiit:
You can also checkin the jars into the repo

AttwellBrian:
Good idea. I'll try that in the morning.

PR #102, opened 2017-11-05
Moving tooling documentation to the Wiki

gergelyorosz:
Following the approach with tutorials, moving the documentation over to
the Wiki as well. The wiki pages are already in place.

In this commit, also removing assets from the android folder.


artman:
That SLA link in the README actually points to the wrong place... Merging this in.

PR #103, opened 2017-11-05
Change the iOS RIB generation template to generate a view by default

gergelyorosz:
This is because of two reasons:
1. A majority RIBs that do have views (both in practice, as well as in the examples)
2. To have the code generation defaults be in line with the Android generator (and the main use case). Confirmed with @sbarow, who has contributed much to the original code generator.


kovpas:
Makes sense, thanks!

PR #106, opened 2017-11-06
Support for Swift Package Manager

### LINKED ISSUE #112: [iOS] Swift Package Manager support
Already implemented in #106. Since SPM doesn't support UIKit imports, I had to `#if os(iOS)` bunch of code, which is not ideal. We have to get back to this once SPM supports iOS builds.

kovpas:
Had to `#if os(iOS)` bunch of UIKit stuff, since SPM doesn't support it yet.

Resolves #112

### DIFF
--- a/ios/RIBs/Classes/LaunchRouter.swift
+++ b/ios/RIBs/Classes/LaunchRouter.swift
@@ -14,6 +14,8 @@
 //  limitations under the License.
 //
 
+#if os(iOS)
+
 import UIKit
 
 /// The root `Router` of an application.
@@ -47,3 +49,5 @@ open class LaunchRouter<InteractorType, ViewControllerType>: ViewableRouter<Inte
         load()
     }
 }
+
+#endif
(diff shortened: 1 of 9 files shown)


ios/RIBs/Classes/LaunchRouter.swift
```diff
@@ -47,3 +49,5 @@ open class LaunchRouter<InteractorType, ViewControllerType>: ViewableRouter<Inte
         load()
     }
 }
+
+#endif
```
artman: Newline please

kovpas: Done

kovpas:
> I don't like the idea that we have to ifdef our code to make it work with Swift package manager. Maybe we should wait for Swift package manager to support platforms before we add support for it?

Yeah, I'm leaving the decision up to you. In other opensource projects, people are usually asking about SPM support. Let's monitor if that's gonna be the case with RIBs and get back to the topic if it is.

alanzeino:
I don't think this is needed. SPM can't even build iOS projects yet.

kovpas:
Yes, good point. Closing the PR.

PR #107, opened 2017-11-06
Fixed URL of RIBs plugin jar

dayanruben:


AttwellBrian:
Thanks for fixing this :)

PR #109, opened 2017-11-06
Renames Android tutorial folders to "tutorial<x>" so that it matches wiki documentation

artman:


AttwellBrian:
I'm fine with this as long as you don't break any of the links in the tutorials.

Didn't the names match before? The `tutorial1-create-a-rib` tutorial was called "Tutorial 1" with a subtitle of "Creating your first RIB".

artman:
The wiki documentation didn't match the folder names.

gergelyorosz:
Having them be in line with the iOS tutorials naming looks good to me.

PR #117, opened 2017-11-07
Fix typo

lucas34:
Change Pesenters to Presenters

### DIFF
--- a/ios/RIBs/Classes/Presenter.swift
+++ b/ios/RIBs/Classes/Presenter.swift
@@ -19,7 +19,7 @@ import Foundation
 /// The base protocol for all `Presenter`s.
 public protocol Presentable: class {}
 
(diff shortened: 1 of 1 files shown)


ios/RIBs/Classes/Presenter.swift
```diff
@@ -19,7 +19,7 @@ import Foundation
 /// The base protocol for all `Presenter`s.
 public protocol Presentable: class {}
 
-/// The base class of all `Pesenters`. A `Pesenters` translates business models into values the corresponding
+/// The base class of all `Presenters`. A `Presenters` translates business models into values the corresponding
```
artman: This should actually be:

The base class of all `Presenter`s. A `Presenter` translates business models into values the corresponding

kovpas:
Thanks for the fix @lucas34!

PR #119, opened 2017-11-08
Replace `SampleInteractable` with correct template on routers without…

oronbz:
… views


sbarow:
Thanks @oronbz can you sign the CLA so we can merge this.

oronbz:
Done @sbarow, thanks!

oronbz:
@sbarow pusing to pod trunk? (:

artman:
This didn't change files that are included by the podspec, so no need to push a new version. I wish we could still run custom installation scripts when a pod is installed to make sure that users have the correct templates, but I think that got deprecated, right?

sbarow:
@artman we could use [prepare_command](https://guides.cocoapods.org/syntax/podspec.html#prepare_command) for this right?

PR #120, opened 2017-11-08
Bump version to 0.9.2

oronbz:
To include latest commit.

Please also push to `pod trunk` afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### Version 0.9.2
+
+* Add file header to iOS template files
+* Add snapshot credentials
+* Fix typos
+* Fix URL of RIBs plugin jar
+
 ### Version 0.1.0
 
 * Initial release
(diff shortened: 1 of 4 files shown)


sbarow:
Thanks @oronbz can you follow the steps in [RELEASING.md](https://github.com/uber/RIBs/blob/master/RELEASING.md) 1-4 and I'll do the rest.

oronbz:
Thanks @sbarow, done for both iOS and Android.

sbarow:
@artman does it make sense to bump the Android version inline with iOS?

artman:
Yes, we need to come up with a strategy and helper script to bump versions on both Android and iOS at the same time. However, I don't see the need to bump the podspec version as #119 or #121 only changed template files.

oronbz:
@artman you're definitely right, I've missed the fact that it's only template changes (one of them was actually mine). Would you like me to close this pull request or just rebase it when there's actual content that will affect gradle/pod?

artman:
Yeah, let's close this out for now. Thanks!

PR #138, opened 2017-11-14
Fix Carthage usage

sbarow:
When fetching & trying to build with Carthage it will fail due to some dangling cocoa pods references as well as not being able to link to `RxSwift`. 

This diff does 5 things:
- Deintegrate cocoa pods 
- Update `FRAMEWORK_SEARCH_PATHS` with path to checked out Carthage build folder.
- Updates `.gitignore` & removes untracked cached files.
- Deletes the `Podfile` 
- Moves the `Cartfile` to the route of the project so dependencies will also be downloaded (`RxSwift `)

Fixes: #149


kovpas:
Why do we need this?

sbarow:
@kovpas yeah, updated the description - tested this locally with `Carthage` & `Cocoapods` and its all working properly.

sbarow:
@artman ping on this. I have made changes to the travis build script and this would change how we do local dev. Let me know if you are happy for me to merge this.

Cartfile
```diff
@@ -0,0 +1 @@
+github "ReactiveX/RxSwift" ~> 4.0
```
artman: Why are we including the cartfile? This is not required, right?

Cartfile.resolved
```diff
@@ -0,0 +1 @@
+github "ReactiveX/RxSwift" "4.0.0"
```
artman: This is not need either.

.travis.yml
```diff
@@ -32,22 +32,19 @@ matrix:
...
         - LANG=en_US.UTF-8
-        - WORKSPACE=RIBs.xcworkspace
         - IOS_SDK=iphonesimulator11.0
 
-      before_install:
-          - gem install cocoapods --no-rdoc --no-ri --no-document --quiet
-
       script:
         - set -o pipefail
+        - carthage update --platform iOS
```
artman: Can we keep this building with pods, too?

.travis.yml
```diff
@@ -32,22 +32,19 @@ matrix:
...
+        - mv Carthage/ ios
```
artman: Not sure what this is doing

ios/Podfile
```diff
@@ -1,10 +0,0 @@
-platform :ios, '8.0'
```
artman: Why did we delete this?

sbarow:
@artman some explanations to the above:

1. Podfile deletion - I did this so local development isn't done with cocoa pods, the reason for this is when someone does a pull request without having done `pod deintegrate` the Carthage implementation will break as there are references to cocoa pod specific files. So this is more of a fail safe.
2. Cartfile - we need that for users whom use Carthage so when `carthage update` is done, `RxSwift` will also be downloaded & built. If we don't do this then users have to add the `RxSwift` dependency themselves to their `Cartfile` - its essentially the same as a `.podspec`
3. Cartfile.resolved - I'll remove it, we shouldn't need that.
4. `- mv Carthage/ ios` because of our folder structure we have to do this. We need the `Cartfile` in the root for the reason mentioned in `2`. When `carthage update` runs it creates a `Carthage/Build` folder with the built frameworks, which needs to be moved to the `ios` folder for `FRAMEWORK_SEARCH_PATHS` to be able to reference the built frameworks.
5. Can we keep building with cocoapods - we would have to add the `Podfile` back which is what we would like to avoid because of `1.`

PR #171, opened 2017-11-29
Added RIB's template option to create a corresponding XIB file

kronik:
Changes:

1. Added another checkbox in a new RIB template
2. This checkbox is active only in case when "owns the corresponding view" is checked
3. Installation script copies "default" RIB files into template with XIB folder


kronik:
I've signed CLA agreement but this tool still shows that it's not signed.

sbarow:
Tested this out locally, looks good on my end. 

cc @neakor @kovpas

PR #180, opened 2017-12-05
Fix tutorials view-less RIB inheritance

neakor:
View-less RIBs should not inherit from `ViewableRouter`. It'll cause false positive leak detection.


neakor:
I checked the wiki, they don't mention inheritance. I assume that's because they were relying on the templates to generate the correct inheritance. But the templates were wrong 😆

sbarow:
🤦‍♂️ thanks Yi.

PR #188, opened 2017-12-14
added PR template

jarinosuke:
to improve community checks 👍 

screenshot


hzsweers:
I think I'd prefer something with a bit more detail. We can leverage comment tags like we do in AutoDispose - https://github.com/uber/AutoDispose/tree/master/.github (make sure to view the raw files, not rendered)

@artman @AttwellBrian thoughts?

jarinosuke:
@hzsweers Thanks for your help!

> I think I'd prefer something with a bit more detail.

Yes, I think so too.

Should I move this file into `.github` directory? I put it to root since [CONTRIBUTING.md](https://github.com/uber/RIBs/blob/master/CONTRIBUTING.md) is there.

hzsweers:
Yeah it does need to be in `.github` for github to pick it up I think

hzsweers:
@jarinosuke if you want to just copy what we have in autodispose, that would be a good start

jarinosuke:
@hzsweers Thanks for your advise! I've copied this https://github.com/uber/AutoDispose/blob/master/.github/PULL_REQUEST_TEMPLATE.md

hzsweers:
Thanks!

PR #196, opened 2017-12-27
Update template to call super with passed in viewController

### LINKED ISSUE #192: Creating a viewless RIB from template doesn't compile
When creating a **viewless** RIB from the template, say for example you call it MyRib, the MyRibRouter class inherits from Router<MyRibInteractable>. The template has the MyRibRouter overriding Router's initializer:
`public init(interactor: InteractorType, viewController: ViewControllerType) { ... }` 

But that's the initializer for a `ViewableRouter`, a `Router's`  initializer is:
 `public init(interactor: InteractorType) { ... } ` 

The code doesn't compile because the initializer does not override a designated initializer from MyRibRouter's superclass. The issue seems to be related to PR #179 Maybe I don't understand those changes?

Steps to reproduce:
1. create a RIB from the template
2. uncheck Owns corresponding view
3. try to compile the code

I can open a PR if desired. Unless these changes were on purpose? @neakor

sbarow:
Fixes #192

### DIFF
--- a/ios/tooling/RIB.xctemplate/Default/___FILEBASENAME___Router.swift
+++ b/ios/tooling/RIB.xctemplate/Default/___FILEBASENAME___Router.swift
@@ -18,7 +18,7 @@ final class ___VARIABLE_productName___Router: Router<___VARIABLE_productName___I
     // TODO: Constructor inject child builder protocols to allow building children.
     override init(interactor: ___VARIABLE_productName___Interactable, viewController: ___VARIABLE_productName___ViewControllable) {
         self.viewController = viewController
-        super.init(interactor: interactor, viewController: viewController)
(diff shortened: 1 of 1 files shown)


ios/tooling/RIB.xctemplate/Default/___FILEBASENAME___Router.swift
```diff
@@ -18,7 +18,7 @@ final class ___VARIABLE_productName___Router: Router<___VARIABLE_productName___I
     // TODO: Constructor inject child builder protocols to allow building children.
```
edrew08: I think you would also have to remove the `override` keyword because you're not overriding a designated initializer from the superclass.

PR #203, opened 2018-01-04
Add Xcode Project Template

edrew08:
Is there any interest in a project template for Xcode? I created a project template that allows a user to create a RIBs app and name the root RIB of their project. It's installed with the same install script used for the file templates. 

This would make it easier for people to get their RIBs project off the ground and running. It would also save user's time from having to create a RIB and change it to conform to `LaunchRouter`, and updating their AppDelegate. After creating a RIBs App, a user would just have to install the RIBs cocoapod (or carthage) and then their project would be ready to go.

There are a couple issues currently:
1. A new template icon for a RIBs App project template would probably be a good idea
2. The variable identifier in the current RIBs templates for the RIB name identifier is `productName`. The same identifier is used by Apple for a project template's Product Name. The problem is that when changing the Root RIB Name when setting up your project, it also changes the bundle identifier because the bundle identifier is using the `productName` identifier. Basically, the problem is that there are duplicate `productName` variable identifiers. One solution would be to update the RIB templates to use a different identifier for the RIB name variable. Thoughts?

screen shot 2018-01-04 at 5 37 28 pm


ios/tooling/Project Templates/RIBs App.xctemplate/TemplateInfo.plist
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Concrete</key>
+	<true/>
+	<key>Kind</key>
+	<string>Xcode.Xcode3.ProjectTemplateUnitKind</string>
+	<key>Identifier</key>
+	<string>com.uber.BaseApplication</string>
```
edrew08: This identifier should probably be something more meaningful but I don't know what your naming conventions are.

ios/tooling/Project Templates/RIBs App.xctemplate/TemplateInfo.plist
```diff
@@ -0,0 +1,85 @@
...
+	</array>
+	<key>Options</key>
+	<array>
+		<dict>
+			<key>Identifier</key>
+			<string>languageChoice</string>
+			<key>Values</key>
+			<array>
+				<string>Swift</string>
+			</array>
```
edrew08: Restricting the language choice to Swift only.

sbarow:
Thanks for this, looks like it could be useful for bootstrapping a project. 

Thoughts @artman @neakor

neakor:
Thanks for the interest. Although I'm not sure if we want to suggest a single particular way to setup a project with RIBs.

edrew08:
True, people would still have the option of creating a regular Xcode project and integrating RIBs how/when they want. My thinking was that every time I start a RIBs project I find myself having to do the same steps over and over again to get everything set up. Maybe this project template pattern doesn't cover every scenario but it at least provides a jumping off point, just like existing Apple Xcode project templates do.

PR #213, opened 2018-02-07
Enable NullAway on memory leaks demo

msridhar:
Bump Error Prone and NullAway versions, and leverage the `-XepExcludedPaths` option to prevent checking of Dagger-generated code.

### DIFF
--- a/android/demos/memory-leaks/build.gradle
+++ b/android/demos/memory-leaks/build.gradle
@@ -59,6 +59,7 @@ dependencies {
     annotationProcessor deps.apt.daggerCompiler
     annotationProcessor project(":libraries:rib-compiler-app")
     annotationProcessor project(":libraries:rib-compiler-test")
+    annotationProcessor deps.apt.nullAway
     compile project(":libraries:rib-android")
     compile deps.external.dagger
     compile deps.external.rxbinding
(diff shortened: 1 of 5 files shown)


msridhar:
@AttwellBrian do we want to enable NullAway anywhere else?

AttwellBrian:
This is probably good enough. It means we can remove the [following quote from the RIBs tutorial](https://github.com/uber/RIBs/wiki/Android-Tutorial-1):

> Let’s try building and installing the app. You’ll notice it crashes because there is more work to do (and because we don't have NullAway configured).

AttwellBrian:
Will the `XepExcludedPaths` be setup thoroughly enough for engineers to work all the way through the first tutorial without hitting NullAway problems?

msridhar:
> Will the `XepExcludedPaths` be setup thoroughly enough for engineers to work all the way through the first tutorial without hitting NullAway problems?

Not sure; will take a look.  I only tweaked the one `build.gradle` file where I saw a comment about NullAway.

msridhar:
@AttwellBrian looks like NullAway is not really wired up anywhere except for where I added it (the memory leak demo).  Do you want it wired up (i.e., added to Gradle build files) for the tutorials as well?

AttwellBrian:
It would be nice if you could wire it up for Tutorial 1 so we can include a NullAway endorsement in that tutorial. But it is up to you whether you think this is worth your time :)

You would have to verify that the tutorial can be completed with NullAway enabled (maybe by testing to see if tutorial 2 can build with nullaway).

msridhar:
Just checking: are we confident that NullAway will actually catch the NPE that turns up in tutorial1? 😄 

Also, when I added Error Prone to tutorial2 I get this error:
```
RIBs/android/tutorials/tutorial2/src/main/java/com/uber/rib/root/logged_out/LoggedOutInteractor.java:31: error: [CheckReturnValue] Ignored return value of method that is annotated with @CheckReturnValue
            .subscribe(new Consumer<String>() {
                      ^
    (see http://errorprone.info/bugpattern/CheckReturnValue)
```

Looks like we're not doing anything with the return of a `subscribe`.  If I disable that check everything compiles (i.e. there are no NullAway issues).  How should the above be addressed?  I think assigning the result to an unused local is enough to get rid of the error.

AttwellBrian:
> Just checking: are we confident that NullAway will actually catch the NPE that turns up in tutorial1? 

It was caught when I originally wrote the tutorial in the mono-repo. The code in question will look like:

```
  @Override
  protected LoggedOutView inflateView(LayoutInflater inflater, ViewGroup parentViewGroup) {
    return null;     // TODO: inflate view here
  }
```

> How should the above be addressed? I think assigning the result to an unused local is enough to get rid of the error.

In this case there is no reason to check the return value? So just disable this check?

msridhar:
Yeah that error should definitely get caught.  I disabled `CheckReturnValue` for tutorial2.  I think this one is ready to merge now assuming it passes on CI.

PR #229, opened 2018-03-30
Remove weak property keyword from protocols

edrew08:
**Description**:
* Removing `weak` from property declarations in protocols. Removed from both the Xcode templates and tutorials.
* Removes the warning in Xcode 9.3, `'weak' should not be applied to a property declaration in a protocol and will be disallowed in future versions`
* https://github.com/apple/swift-evolution/blob/master/proposals/0186-remove-ownership-keyword-support-in-protocols.md
* Maybe a couple changes needed on the wiki? I don't know how to edit the wiki.

**Related issue(s)**: N/A


sbarow:
Thanks @edrew08

PR #243, opened 2018-04-21
Fixed typo on POM artifact id for rib-debug-utils

jorgecoca:
**Description**: Fixed typo on the POM artifact id for the Android debug utils


AttwellBrian:
Thanks @jorgecoca !

PR #244, opened 2018-04-21
bugfix: expectDeallocate and expectViewControllerDisappear ,Added dup…

FreddyZeng:
**Description**:
expectDeallocate and expectViewControllerDisappear,  added  duplicate key for iOS


neakor:
Two expectations should be fine. If there's no leak, both expectations should just pass.

Issue #281, opened 2018-10-11
[iOS] LaunchRouter should not come with ViewController


dangthaison91:
In other way, should **LaunchRouter** come with **UIWindow** only.  When it attaches child RIBs, child's **ViewController** will be set as **rootViewController**?

This will remove **unused** RootViewController and give it's child flexibility.

What do you think? I'm willing to make a PR.

sbarow:
Can you give an example of when a RootViewController is **unused**? In the case of having an app that has a logged in state its useful to have a RootViewController that can embed the logged in or logged out controller.

dangthaison91:
@sbarow We want to attach many types of ViewControleller like UITabbarViewController/UINavigationController... and we want make it as **window.rootViewController**. So when you in logged in state, you can embed LoggedInViewController as rootViewController of UIWindow.

dangthaison91:
At the moment, LaunchRouter will have RootViewController and never be detached, always be window.rootViewController. So we can only present/push child ViewController from root.  We want something like this:

**RootRouter:**
```
func routeToOnboarding() {
    window.rootViewController = OnboardingViewController
}

func routeToHomeTabbar() {
    window.rootViewController = HomeTabbarViewController
}

func routeToLoggedOut() {
    window.rootViewController = LoggedOutNavigationViewController
}
```

If there are other way to do this, please tell me?

sbarow:
You could just do:

```
// This property holds the UIWindow object in the windows array that is most recently sent the makeKeyAndVisible() message.
private var window: UIWindow? {
    return UIApplication.shared.keyWindow
}

func routeToOnboarding() {
    window?.rootViewController = OnboardingViewController
}

func routeToHomeTabbar() {
    window?.rootViewController = HomeTabbarViewController
}

func routeToLoggedOut() {
    window?.rootViewController = LoggedOutNavigationViewController
}
```

dangthaison91:
@sbarow Thanks, but I didn't find a place to put this code. RIB should know their child but not the parent? So they should not change the view hierarchy from the root. I prefer to introduce a Router that will hold a UIWindow reference let developers decide how to use it.

sbarow:
I don't understand what you mean by 
> I didn't find a place to put this code

Regardless of if you have a Router that has a reference to the UIWindow or not you are going to have to be able to call the methods needed to replace the window. 

The above code should work for you, it should live in your "Root" router, your children would communicate to its parent that they want to transition to one of the routes through the listener interface.

[...]

PR #284, opened 2018-10-12
[WIP] LaunchRouter now will come with UIWindow

dangthaison91:
This PR related to issue #281.

**Description**:
This will replace UIViewController by UIWindow in LaunchRouter. This will help we create View Hierarchy as before: SignIn/Home/Main... will be set as UIWindow.rootViewController. They will no longer be present from RootViewController.

In a short way, RootViewController now becomes a UIWindow.

### DIFF
--- a/ios/RIBs/Classes/LaunchRouter.swift
+++ b/ios/RIBs/Classes/LaunchRouter.swift
@@ -34,7 +34,7 @@ open class LaunchRouter<InteractorType>: Router<InteractorType>, LaunchRouting {
     /// - parameter window: The corresponding `UIWindow` of this `Router`.
     private weak var window: UIWindow?
 
-    public override init(interactor: InteractorType, window: UIWindow) {
+    public init(interactor: InteractorType, window: UIWindow) {
         self.window = window
         window.makeKeyAndVisible()
         super.init(interactor: interactor)
--- a/ios/RIBs/Classes/LaunchRouter.swift
+++ b/ios/RIBs/Classes/LaunchRouter.swift
@@ -17,33 +17,36 @@
 import UIKit
 
 /// The root `Router` of an application.
-public protocol LaunchRouting: ViewableRouting {
+public protocol LaunchRouting: Routing {
 
     /// Launches the router tree.
     ///
     /// - parameter window: The application window to launch from.
-    func launch(from window: UIWindow)
+    func show(viewController: UIViewController)
 }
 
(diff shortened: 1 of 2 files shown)


neakor:
I'm not sure I follow why we are making this change? The original setup is to allow the creation of `UIWindow` independent of RIBs. `UIWindow` is strictly a UIKit concept that's outside the domain of RIBs. I would rather not push this logic into the router.

neakor (changes requested):
I'm not sure I follow why we are making this change? The original setup is to allow the creation of `UIWindow` independent of RIBs. `UIWindow` is strictly a UIKit concept that's outside the domain of RIBs. I would rather not push this logic into the router.

neakor:
@dangthaison91 instead of modifying the existing one, can we add a new one?

dangthaison91:
@neakor I'm not sure new one like `WindowRouter` or something else will make sense? Introduce new Router type could make RIB framework a little bit more complex? Could you share your ideas?

How about store UIWindow as a property inside LaunchRouter and provide a `replaceRootViewController(:)`?

neakor:
The logic of replacing the root view controller doesn't seem related to Router or RIB to me. Can we just do this logic outside of the `LaunchRouter`?

ermik:
`rootViewController` is a platform convention, it's part of how storyboards work or it is implemented by the developer themselves. If we agree RIBs have no business creating UIWindow or rootViewController, it is important to say `launchFrom` feels like a natural of other RIBs idioms. It fits.

Also, this PR doesn't touch the tutorials code and has some comments that it has missed in the code.

dangthaison91:
> The logic of replacing the root view controller doesn't seem related to Router or RIB to me.
IMHO, we are configuring rootViewController inside LaunchRouter, which is against your statement. Should we change `LaunchRouter` to have a `UIWindow` instead of UIViewController?  Then we leave user decide how to configure their view hierarchy inside their `LaunchRouter` subclass?

Another option is introducing `WindowRouter` but I think we will have some questions about  the name and the existing of `LaunchRouter`

sbarow:
This PR doesn't make sense, please use the solution suggested in #281

dangthaison91:
I ended up by creating WindowRouter for internal use only instead of modifying LaunchRouter. Need time to see how it works also consider other approaches.

neakor:
@dangthaison91 I'll close this PR for now. Please let us know in the future. Thanks for the contribution!

Issue #287, opened 2018-10-15
[iOS] Replace RxSwift Variable with BehaviorRelay


dangthaison91:
From RxSwift 4.0, Variable is marked as DEPRECATED and will be removed soon. We should update `LeadDetector` to using `BehaviorRelay`.

Caused BehaviorRelay is a part of RxCocoa, we have to add RxCocoa as a dependency in podspec. Another option is to pull BehaviorRelay's code from RxSwift source only and add it to current Uber RIBs.
I voted for option 2 because current BehaviorRelay code is simple and seem not to change in future.

alexvbush:
The iOS implementation now has a new home https://github.com/uber/RIBs-iOS/

This issue was migrated to https://github.com/uber/RIBs-iOS/issues/38

We can close this thread.

PR #289, opened 2018-10-22
Replace Variable by BehaviorSubject

dangthaison91:
This will remove using deprecated Variable by using BehaviorSubject. This is related to #287 
I chose to use `BehaviorSubject` instead of introducing `BehaviorRelay` although it is safer to use.


sbarow:
Should we wait until this is official deprecated in `RxSwift 5.0`. As per the documentation.

> This concept will be deprecated from RxSwift but offical migration path hasn't been decided yet.

I'd prefer to wait for the official migration path to be decided before we do this.

neakor:
It might be better to switch these to use `AtomicInt` and `AtomicBool` from https://github.com/uber/swift-concurrency instead.

sbarow:
This has been fixed with #322

sbarow:
> It might be better to switch these to use `AtomicInt` and `AtomicBool` from https://github.com/uber/swift-concurrency instead.

This is a good idea.

PR #292, opened 2018-10-30
[ios/tooling] install-xcode-template respects the `xcode-select`ed Xcode

AndrewSB:
I'm not a bash pro, please patch this to be less terrible if it is. I tested on my computer and this correctly output the template path relative to my `xcode-select`ed Xcode


AndrewSB:
This is invalid, I was wrong: All templates are stored in the directory inside ~/Library, they aren't Xcode specific. Apologies for the noise.

PR #305, opened 2018-12-08
Remove use of soon to be deprecated Variable

ccwasden:
**Description**:
From RxSwift 4.0, Variable is marked as DEPRECATED and will be removed soon. This change removes the use of `Variable` in `LeakDetector` and instead uses an ivar which pushes to a `ReplaySubject` in `didSet`. The behavior with this change should match the current `Variable` approach exactly.

**Related issue(s)**:
https://github.com/uber/RIBs/issues/287

There are no unit tests for `LeakDetector`.

### DIFF
--- a/ios/RIBs/Classes/LeakDetector/LeakDetector.swift
+++ b/ios/RIBs/Classes/LeakDetector/LeakDetector.swift
@@ -59,7 +59,7 @@ public class LeakDetector {
     /// The status changes between InProgress and DidComplete as units register for new detections, cancel existing
     /// detections, and existing detections complete.
     public var status: Observable<LeakDetectionStatus> {
-        return expectationCount
+        return expectationCountSubject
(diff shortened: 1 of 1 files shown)


ccwasden:
Though we may not merge until there is an official migration path as mentioned in this PR https://github.com/uber/RIBs/pull/289 I propose this PR as a better alternative to 289

ermik:
@ccwasden do you have an idea of why Android CI failed?

ccwasden:
@ermik No clue. If you see my diff it's only swift changes. I just committed a minor change to try and force a CI re-run. I assume there aren't typically intermittent failures on android CI?

ccwasden:
Android is failing CI on master: https://travis-ci.org/uber/RIBs/branches

sbarow:
This has been fixed with #322

PR #309, opened 2019-01-08
Add default implementation to LaunchRouting

mohamede1945:
**Description**:
Add default implementation to LaunchRouting. By doing this, we can use it for custom routers and not restricted to use LaunchRouter.
Mainly I moved the implementation from LaunchRouter to be a default implementation in for LaunchRouting.

### DIFF
--- a/ios/RIBs/Classes/LaunchRouter.swift
+++ b/ios/RIBs/Classes/LaunchRouter.swift
@@ -25,6 +25,20 @@ public protocol LaunchRouting: ViewableRouting {
     func launch(from window: UIWindow)
 }
 
+public extension LaunchRouting {
+
+    /// Launches the router tree.
+    ///
+    /// - parameter window: The window to launch the router tree in.
+    public func launch(from window: UIWindow) {
+        window.rootViewController = viewControllable.uiviewController
+        window.makeKeyAndVisible()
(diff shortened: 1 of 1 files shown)


mohamede1945:
CI Failing on Android: JDK oracle:jdk8 job. Xcode one passed.

ermik:
I'm sure there was a reason `LaunchRouter` was meant to be final and required. This PR seems like a departure from strict implementation procedure that makes RIBs such an attractive _mindset_. In my opinion these changes, while sensible, belong in a forked repository.

mohamede1945:
`LaunchRouting` is a public protocol that can be adapted by any router. So there is no strict procedure lifted or whatsoever.

If I have a tree structure of RIBs, one could suppose we can make any of the node RIBs the entry point of the app that by just implementing `LaunchRouting` protocol which we can currently, but we will have to implement `launchFromWindow` in each router that could possibly be a `LaunchRouting`. Having a default implementation for the `LaunchRouting` would prevent direct access by these routers to 
```
interactable.activate()
load()
```
Which as per docs they are internal to the framework and shouldn't be used directly by apps.

So in contrast to your claim, this PR is in fact to minimze access to stricted areas of RIBs.

PR #311, opened 2019-02-01
Update the code to be compatible with Swift 5.0

ibrahimkteish:
**Description**:
Migrate the code to Swift 5.0

**Related issue(s)**:
NA

Issue #274, opened 2018-09-11
[iOS] Regression in Component.shared behavior in Xcode 10


dzzh:
Because of https://bugs.swift.org/browse/SR-8704, there's a regression in `Component.shared` behavior. In Xcode 10, it always returns nil when invoked inside functions that have optional return types.

D2055355

PR #321, opened 2019-05-02
Upgrade iOS RIBs Project to Swift 5

JUSTINMKAUFMAN:
**Description**:
Most iOS projects remain in close synchronization with the most recent Swift/Xcode release version. Because the iOS side of the RIBs library currently only builds for Swift 4, iOS projects are forced to fork the official RIBs library and upgrade the project to Swift 5 manually.

This PR upgrades the iOS project - along with related dependencies and repository files - to Swift 5.

**Related issue(s)**:
N/A

I do not believe the minor changes proposed in this PR require any new unit tests (though I am happy to add some if I am told what such tests would look like).


ios/RIBs/Classes/DI/Component.swift
```diff
@@ -47,14 +47,6 @@ open class Component<DependencyType>: Dependency {
             lock.unlock()
         }
 
-        #if swift(>=5.0)
```
JUSTINMKAUFMAN: I'm assuming this was added with the idea that SR-8704 would be resolved with the release of Swift 5. It's unclear to me how that particular Swift bug was addressed (it is now closed), but regardless, the behavior you addressed in the #else block is still required. Thus, I have removed the #if swift(>=5.0) branch.

ermik: @dzzh this is something you had the chance to catch in #274 and address with that very helpful test. I am eager to hear anything you have to say and suggest any further improvements you think are appropriate here.

ermik:
This PR also updates the Travis build setup and with the fix in c2bd5a981858812dee716a34d1725755d42094a9 tests now [pass](https://travis-ci.org/uber/RIBs/jobs/527565408) for iOS. We are looking into the situation with the continuing Android failure — it looks like a CI issue due to unpinned dependency. But while it is a nuisance any improvements should remain outside the scope of this PR.

@ibrahimkteish could you look at the changes we've made and see if you can close #311 (your PR for Swift 5) in favor of the work done here?

Requesting review from people I know to have previously reviewed iOS PRs: @sbarow @neakor @artman.

ios/RIBs/Classes/DI/Component.swift
neakor: I think this is fine for now. I'd like to remove DI completely from this repo, since this is independent of RIBs.

JUSTINMKAUFMAN:
> Thanks much for this. Internally at Uber we are not at Swift 5 yet due to significant Xcode 10.2 compile-time speed regression.

@neakor That's interesting to hear. We actually realized some compile-time performance improvements, though I imagine your codebase is a beast compared to ours (size-wise, that is) :).

In any case, how do we get this merged given the Travis CI build failure? 

cc: @ermik

neakor:
I'm merging this since the failure is within Android and this PR didn't touch Android in anyway.

neakor:
@Leland-Takamine can you please take a look at the Android failure?

ermik:
@Leland-Takamine I tracked it down to the `tools` dependency not being pinned (basically Travis environment is not immutable over long periods of time). travis-ci/travis-ci#5049 talks about the history of this but does not provide a solution.

ermik:
Do we know how to make a new release for Cocoapods / Carthage? Honestly, I just don't want any more Swift 5 PRs being spawned 🙀 because people can't access these changes (unless they pin to a commit hash of course).

PR #322, opened 2019-05-06
Compatible RxSwift 5

daidongon:
**Description**:
- Compatible RxSwift 5.x
  - migrate from `Variable` to `BehaviorRelay`
  - Replace `TimeInterval` to `DispatchTimeInterval` in `Executor`
- Fix Dependency
  - Cartfile
  - RIBs.podspec

**Related issue(s)**:
N/A

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "ReactiveX/RxSwift" ~> 5.0.1
+github "ReactiveX/RxSwift" ~> 5.0
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "ReactiveX/RxSwift" ~> 4.5.0
+github "ReactiveX/RxSwift" ~> 5.0.1
--- a/RIBs.podspec
+++ b/RIBs.podspec
@@ -12,4 +12,5 @@ RIBs is the cross-platform architecture behind many mobile apps at Uber. This ar
   s.ios.deployment_target = '8.0'
   s.source_files = 'ios/RIBs/Classes/**/*'
   s.dependency 'RxSwift', '~> 5.0'
+  s.dependency 'RxRelay', '~> 5.0'
 end
--- a/RIBs.podspec
+++ b/RIBs.podspec
@@ -11,5 +11,5 @@ RIBs is the cross-platform architecture behind many mobile apps at Uber. This ar
   s.source           = { :git => 'https://github.com/uber/RIBs.git', :tag => 'v' + s.version.to_s }
   s.ios.deployment_target = '8.0'
   s.source_files = 'ios/RIBs/Classes/**/*'
-  s.dependency 'RxSwift', '~> 4.5.0'
+  s.dependency 'RxSwift', '~> 5.0.0'
 end
--- a/RIBs.podspec
+++ b/RIBs.podspec
@@ -11,5 +11,5 @@ RIBs is the cross-platform architecture behind many mobile apps at Uber. This ar
   s.source           = { :git => 'https://github.com/uber/RIBs.git', :tag => 'v' + s.version.to_s }
   s.ios.deployment_target = '8.0'
   s.source_files = 'ios/RIBs/Classes/**/*'
-  s.dependency 'RxSwift', '~> 5.0.1'
+  s.dependency 'RxSwift', '~> 5.0'
 end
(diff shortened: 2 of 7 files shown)


avitruk:
Hi all, 
@daidongon found bug in `Executor`. Line 51. 
```swift
Double(maxFrameDuration / 1000) // always return 0
```

should be :

```swift
Double(maxFrameDuration) / 1000
```

@ermik When we can expect RxSwift 5 compatibility release?

ios/RIBs/Classes/LeakDetector/Executor.swift
```diff
@@ -46,7 +46,7 @@ public class Executor {
 
                 // If we did drop frame, we under-count the frame duration, which is fine. It
                 // just means the logic is performed slightly later.
-                let boundedElapsedTime = min(trueElapsedTime, maxFrameDuration)
+                let boundedElapsedTime = min(trueElapsedTime, Double(maxFrameDuration / 1000))
```
avitruk: Hi all, 
@daidongon found bug:
```swift
Double(maxFrameDuration / 1000) // always return 0
```

should be :

```swift
Double(maxFrameDuration) / 1000
```

@ermik When we can expect RxSwift 5 compatibility release?

daidongon: @avitruk Thanks for reviewing!!  I'll fix it.

ermik: I don't have any powers as to releases, but together we will make it happen.

dehrom: Time to increase RxSwift version :)

RIBs.podspec
```diff
@@ -11,5 +11,5 @@ RIBs is the cross-platform architecture behind many mobile apps at Uber. This ar
   s.source           = { :git => 'https://github.com/uber/RIBs.git', :tag => 'v' + s.version.to_s }
   s.ios.deployment_target = '8.0'
   s.source_files = 'ios/RIBs/Classes/**/*'
-  s.dependency 'RxSwift', '~> 4.5.0'
+  s.dependency 'RxSwift', '~> 5.0.1'
```
dsrees: I think `~> 5.0` is sufficient or else this could easily cause conflicts with other Pods in a project that use, say, 5.0.2 or 5.1 when those versions are released

dehrom:
Please smb review it)

RIBs.podspec
```diff
@@ -11,5 +11,5 @@ RIBs is the cross-platform architecture behind many mobile apps at Uber. This ar
   s.source           = { :git => 'https://github.com/uber/RIBs.git', :tag => 'v' + s.version.to_s }
   s.ios.deployment_target = '8.0'
   s.source_files = 'ios/RIBs/Classes/**/*'
-  s.dependency 'RxSwift', '~> 4.5.0'
+  s.dependency 'RxSwift', '~> 5.0'
```
bmwalters: It looks like RIBs needs to explicitly depend on `RxRelay` as well.

Cartfile
```diff
@@ -1 +1 @@
-github "ReactiveX/RxSwift" ~> 4.5.0
\ No newline at end of file
+github "ReactiveX/RxSwift" ~> 5.0.1
```
bmwalters: Should the version be relaxed here as it was in the `podspec`?

RIBs.podspec
ermik: Good catch. Is there a specialized testing facility for Cocoapods that we could use to verify that specs are complete or is it just "try to xcodebuild"?

bmwalters: Yeah, you can run `pod lib lint`.
https://stackoverflow.com/a/32328089

bmwalters:
@daidongon Looks like you still need to sign the CLA [linked above](https://github.com/uber/RIBs/pull/322#issuecomment-489668443). Could this then be reviewed by @neakor?

dehrom:
I want this merged in bounds of my lifetime 🙂

smartappdeveloop:
Is anybody from Uber here? Please, merge this pull request, we are waiting for it since 6 of may. There is also a problem with deprecated UIWebView an RxSwift, for more details https://github.com/ReactiveX/RxSwift/issues/2055 .
So, we need to use latest RxSwift in RIBs now, and even more the version with fixed UIWebView.
Very sad to see that last RIBs release was made 1.5 years ago :(

sbarow:
Thanks for this PR @daidongon sorry for the long review times.

PR #327, opened 2019-05-20
Add new builder classes compatible with Needle

neakor:
Since Uber internally has fully migrated to [Needle](https://github.com/uber/needle/), these base builder classes were introduced to make RIBs more compatible.


neakor:
Android is failing. Unrelated to this PR.

PR #356, opened 2019-11-02
Update Podfile

Donald90:
**Description**:
There's a conflict between _tutorial1_ and _RIB_ because they depend on two different versions of RxSwift.
_RIBs_ pod has a dependency on `RxSwift ~> 5.0` while _tutorial1_ has a dependency on `RxCocoa ~> 4.0.0` that has RxSwift as subdependecy.


kovpas:
I believe this one is updated to 5.1 with #370 
Thanks for the PR, @Donald90!

PR #357, opened 2019-11-04
Support for Swift Package Manager

### LINKED ISSUE #112: [iOS] Swift Package Manager support

nick-potts:
Adds support for the SPM now that you can specify a platform in the `Package.swift`.

This will require the project adds a new tag/release for v0.9.3.

I've added a couple imports so that it builds when using iOS frameworks - ``swift build -Xswiftc "-sdk" -Xswiftc "`xcrun --sdk iphonesimulator --show-sdk-path\`" -Xswiftc "-target" -Xswiftc "x86_64-apple-ios13.0-simulator"``

Resolves #112 #106

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,6 +117,7 @@ xcuserdata/
 
 # Swift Package Manager
 .build/
+Package.resolved
 
 # CocoaPods
 Pods/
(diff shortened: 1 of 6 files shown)


nick-potts:
closed in favour of #359

PR #364, opened 2019-12-16
Update Readme

jeffxchu:
Replace the deprecated gradle dependency handlers i.e. `compile` and `testCompile` with `implementation` and `testImplementation`.


jbarr21:
Already fixed

PR #369, opened 2020-02-18
Update iOS tutorials to use pods for Rx 5 / SnapKit 5

anagoten:
**Description**:
iOS tutorials fail to build against RIBs

This patch updates all tutorial Podfile(s) to use Rx 5 and SnapKit 5.
SnapKit5 requires bumping ios platform to 10.

I see that #356 already exists but it only updates the Podfile for tutorial1 and does not update SnapKit.

**Related issue(s)**:

### DIFF
--- a/ios/tutorials/tutorial1/Podfile
+++ b/ios/tutorials/tutorial1/Podfile
@@ -1,10 +1,10 @@
-platform :ios, '9.0'
+platform :ios, '10.0'
 
 use_frameworks!
 inhibit_all_warnings!
 
 target 'TicTacToe' do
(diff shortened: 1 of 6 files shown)


dautermann:
@artman doesn't appear to be part of Uber anymore.  Would be super nice if somebody at uber could step up and take ownership of this repo and review/merge these PR's.  As for the travis build failing, looks like a `GITHUB_ACCESS_TOKEN` environment variable needs to be added to bypass the "`Maximum number of login attempts`" error that appears in the Travis log when it attempts to download RxSwift.  [More info here](https://github.com/Carthage/Carthage/issues/1236)

PR #370, opened 2020-03-13
Compatible with RxSwift 5.1

imairi:
**Description**:
RxSwift 5.1 removes UIWebView Reactive Extensions due to [Apple's hard deprecation, starting April 2020](https://developer.apple.com/news/?id=12232019b) .

RIBs also should be compatible with RxSwift 5.1.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "ReactiveX/RxSwift" ~> 5.0
+github "ReactiveX/RxSwift" ~> 5.1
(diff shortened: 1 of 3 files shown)


kovpas:
Could you please rebase and update the version in `Package.swift` as well?

imairi:
@kovpas I updated RxSwift version for SPM.

kovpas:
Looks good to me. Thanks for the contribution @imairi! 👍

Issue #392, opened 2021-02-12
[Android] Align internal and open-source versions of RIBs


jbarr21:
Breaking changes required in order to reconcile the differences with our internal version of RIBs.
- programmatic changes
- artifact renames
- minimal behavioral changes

jbarr21:
Addressed with [v0.10.1](https://github.com/uber/RIBs/releases/tag/v0.10.0)

PR #394, opened 2021-02-13
Align rib-workflow with internal version

jbarr21:
This PR aligns the `rib-workflow` module with our internal version as part of #392. The changes consist of:
- one API change in `Step#onStep` which now passes in the original step as the first parameter
- migrating anonymous inner classes to lambdas
- reordering of imports
- whitespace 

Also included in this change is the rename of the `rib-workflow-test-utils` module to `rib-workflow-test` for consistency with other test util modules.


android/libraries/rib-workflow-test/src/main/java/com/uber/rib/workflow/core/StepTester.java
```diff
@@ -13,13 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
```
ShaishavGandhi, line 16: We could consider adding Spotless to make sure we have consistent spacing and formatting.

ShaishavGandhi (approved):

android/libraries/rib-workflow-test/src/main/java/com/uber/rib/workflow/core/StepTester.java
jbarr21: Yeah, totally makes sense & and I'd like to do that once done with this, but for now, I just want the internal & external versions to match exactly (since internal doesn't run Spotless) so that the remaining differences are more clearly visible.

PR #403, opened 2021-02-24
Add fakes for RIB classes

jbarr21:
Router does not need to be concrete and was only such because iOS does not have abstract classes. Create fakes for easier testing without mocks. #392

### DIFF
--- a/android/libraries/rib-base/build.gradle
+++ b/android/libraries/rib-base/build.gradle
@@ -49,6 +49,9 @@ dependencies {
     testCompile deps.test.junit
     testCompile deps.test.mockito
     testCompile deps.test.truth
+    testCompile(project(":libraries:rib-test")) {
+        transitive = false
+    }
 }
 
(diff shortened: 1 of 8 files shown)


ShaishavGandhi (commented):
Thoughts on calling it `Test*` instead of `Fake*`? `androidx`, `autodispose` etc all have Test prefixes instead of Fake. 

Is it because of alignment with internal version?

jbarr21:
> Thoughts on calling it `Test*` instead of `Fake*`? `androidx`, `autodispose` etc all have Test prefixes instead of Fake.
> 
> Is it because of alignment with internal version?

I left it as FakeThing since that’s what it was internal & its usually better to be explicit. Naming w/ a Fake prefix clearly notes it as a Fake vs a TestThing which could be misunderstood as a mock

ShaishavGandhi (approved):

PR #417, opened 2021-03-25
Add integration with japicmp

### LINKED ISSUE #416: Setup japicmp
Setup japicmp for the migration to Kotlin to catch API compatibility issues
https://github.com/melix/japicmp-gradle-plugin

rafaeltoledo:
May close #416.

Some points to highlight:
- To create a baseline config for comparison, I needed to bump the AGP version.
- I noticed that we do not have the current stable version anywhere (needed to fetch the latest published artifacts to use in the comparison), so I defined it in `gradle.properties`

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -29,6 +29,7 @@ buildscript {
         classpath deps.build.gradlePlugins.android
         classpath deps.build.gradlePlugins.errorprone
         classpath deps.build.gradlePlugins.spotless
+        classpath deps.build.gradlePlugins.japicmp
     }
 }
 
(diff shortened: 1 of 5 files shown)


android/build.gradle
```diff
@@ -95,6 +96,31 @@ subprojects {
     boolean isSample = project.path.startsWith(":demos:") || project.path.startsWith(":tutorials:")
     boolean usesErrorProne = project.name in useErrorProneProjects
 
+    if (isAndroidLibrary) {
```
jbarr21, line 99: we would want to run this for java libraries, as well, since we have some rib modules that are non-android like rib-base

android/gradle/dependencies.gradle
```diff
@@ -64,11 +65,12 @@ def build = [
     errorProneTestHelpers: "com.google.errorprone:error_prone_test_helpers:${versions.errorProne}",
     nullAway: 'com.uber.nullaway:nullaway:0.9.0',
     gradlePlugins: [
-        android: 'com.android.tools.build:gradle:3.6.2',
+        android: 'com.android.tools.build:gradle:4.1.3',
```
jbarr21, line 68: we should probably not bump this in order to keep behavior closer to that of the internal monorepo

android/build.gradle
```diff
@@ -95,6 +96,31 @@ subprojects {
...
+        dependencies {
+            baseline("com.uber.rib:${project.name}:${LAST_VERSION_NAME}") {
+                transitive = false
+                force = true
+            }
+        }
+
+        task japicmp(type: me.champeau.gradle.japicmp.JapicmpTask, dependsOn: 'assemble') {
+            oldClasspath = configurations.baseline.collect { zipTree(it) }.first().filter { it == 'classes.jar'}
+            newClasspath = zipTree(file("$buildDir/outputs/aar/${project.name}-release.aar")).filter { it == 'classes.jar'}
```
jbarr21, line 115: ziptree looks neat, but couldn't get it to work w/ AARs on the older version of AGP so i had to stick w/ my unpackAar task

jbarr21 (changes requested):
thank you for the contribution, but this is already implemented in PR 418

rafaeltoledo:
No problem :)

Issue #412, opened 2021-03-23
Convert RIBs to Kotlin


jbarr21:
Perform a straight convert to Kotlin without API changes before making improvements

ShaishavGandhi:
We can integrate with [japicmp](https://github.com/melix/japicmp-gradle-plugin) to make sure we're API compatible

jbarr21:
> We can integrate with [japicmp](https://github.com/melix/japicmp-gradle-plugin) to make sure we're API compatible

Good suggestion! #416

PR #433, opened 2021-04-14
Convert rib-android to Kotlin

jbarr21:
Migrates the `rib-android` module to Kotlin as part of #412


android/gradle.properties
```diff
@@ -26,3 +26,4 @@ POM_DEVELOPER_ID=uber
 POM_DEVELOPER_NAME=Uber Technologies
 android.useAndroidX=true
 android.enableJetifier=true
+org.gradle.jvmargs=-Xmx4096m
```
tyvsmith, line 29: We were just squeaking by before or leaking somewhere?

tyvsmith (approved):
Approved assuming the xmx bump is valid.

android/gradle.properties
jbarr21: Squeaking by locally. 2nd run would pass due to cached tasks

PR #434, opened 2021-04-16
Convert rib-base to Kotlin

jbarr21:
Migrates the `rib-base` module to Kotlin as part of #412


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
```diff
@@ -0,0 +1,272 @@
...
+import com.uber.rib.core.Rib.Companion.configuration
+import java.util.Locale
+import java.util.concurrent.CopyOnWriteArrayList
+
+/**
+ * Responsible for handling the addition and removal of children routers.
+ *
+ * @param <I> type of interactor this router routes.
+ */
+abstract class Router<I : Any> protected constructor(
```
jbarr21, line 31: Kotlin enforces the finite bounds restriction which prevents cyclical types. this drops Interactor to the looser non-null `Any`. a future improvement could be to add an interface that only `Interactor` implements and change the type to that. that would be similar to [what iOS does](https://github.com/uber/RIBs/blob/80e6cea7bdb7c06a7b403b7cdfe2dfa7f1138109/ios/RIBs/Classes/Router.swift#L76)

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
```diff
@@ -0,0 +1,272 @@
...
+   * its class name to namespace its saved instance state [Bundle] object.
+   *
+   *
+   * If you have multiple children of the same class, use [Router.attachChild] to specify a custom tag.
+   *
+   * @param childRouter the [Router] to be attached.
+   */
+  @JvmOverloads
+  @MainThread
+  internal fun attachChildInternal(childRouter: Router<*>, tag: String = childRouter.javaClass.name) {
```
jbarr21, line 102: Kotlin does not have package-protected equivalent, so I created an internal method that friend modules can use and marked the other method protected

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * The base implementation for all [Interactor]s.
+ *
+ * @param <P> the type of [Presenter].
+ * @param <R> the type of [Router].
+ */
+abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvider<InteractorEvent> {
+  @Inject
+  lateinit var injectedPresenter: P
```
jbarr21, line 38: Dagger can't inject to a nullable type, but the presenter may also be passed in via constructor injection so this is split out into 2 fields

android/build.gradle
```diff
@@ -149,8 +152,13 @@ subprojects {
                     "-Xjvm-default=all"
             ]
             if (project.name in moduleFriends.keySet()) {
-                def outputJarPath = "build/intermediates/compile_library_classes/release/classes.jar"
-                extraCompilerArgs.add("-Xfriend-paths=${project.rootDir}/libraries/${moduleFriends[project.name]}/$outputJarPath")
+                def friendName = moduleFriends[project.name]
```
tyvsmith, line 155: This would be a small and handy Gradle plugin we could provide to use on the modules themselves.

android/libraries/rib-android/src/test/kotlin/com/uber/rib/core/RibActivityTest.kt
```diff
@@ -192,24 +192,23 @@ class RibActivityTest {
       val component: InteractorComponent<ViewPresenter<*>, *> = mock {
         on { presenter() } doReturn(presenter)
       }
-      return EmptyRouter(view, TestInteractor(), component)
+      return EmptyRouter(view, TestInteractor(presenter), component)
+//      return FakeViewRouter<View, TestInteractor>(testInteractor, component)
```
tyvsmith, line 196: delete

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
tyvsmith: Please creating a tracking task for that.

android/libraries/rib-base/src/test/kotlin/android/os/Bundle.kt
```diff
@@ -13,27 +13,27 @@
  * See the License for the specific language governing permissions and
```
tyvsmith, line 13: Weird that it marks InteracotrAndViewModule.java -> Bundle.kt

android/libraries/rib-base/src/test/kotlin/android/os/Bundle.kt
```diff
@@ -13,27 +13,27 @@
...
  * limitations under the License.
  */
-package com.uber.rib.core;
+package android.os
 
-import android.view.View;
+import java.util.HashMap
```
tyvsmith, line 18: Kotlin collections.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/BasicInteractor.kt
```diff
@@ -13,15 +13,14 @@
...
+ * [Interactor] that doesn't rely on field injection.
+ *
+ * @param <P> the type of [Presenter].
+ * @param <R> the type of [Router].
  */
-public enum WorkerEvent {
-  START,
-  STOP
-}
+abstract class BasicInteractor<P : Any, R : Router<*>> protected constructor(
```
tyvsmith, line 24: P: Any or P : Presenter

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -0,0 +1,170 @@
...
+import io.reactivex.Observable
+import javax.inject.Inject
+
+/**
+ * The base implementation for all [Interactor]s.
+ *
+ * @param <P> the type of [Presenter].
+ * @param <R> the type of [Router].
+ */
+abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvider<InteractorEvent> {
```
tyvsmith, line 36: P : Presenter

android/build.gradle
jbarr21: i thought about this, but its very undocumented and prone to breakage over different versions of AGP, Gradle, and Kotlin so I don't think its worth supporting right now (created #436 to track the enhancement if others are interested)

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
jbarr21: #435

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/BasicInteractor.kt
jbarr21: `P : Any` right now since in Java it is currently unbounded

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
jbarr21: `P : Any` right now since in Java it is currently unbounded

android/libraries/rib-android/src/test/kotlin/com/uber/rib/core/RibActivityTest.kt
jbarr21: 3f6b130

android/libraries/rib-base/src/test/kotlin/android/os/Bundle.kt
jbarr21: 3f6b130

tyvsmith (approved):
🚀

Issue #437, opened 2021-04-27
Cannot push events to the callbacksRelay from child class


hamza1592:
## Expected Behavior

We should be able to send events to the the `callbacksRelay` in `RibActivity` from the child classes that inherit from `RibActivity`

## Current Behavior

Interaction to `callbacksRelay` is only one way (can listen to it) using the `public Observable<ActivityCallbackEvent> callbacks()` function but since the `Relay` is hidden we aren't able to send events to it.

## Possible Solution

Create a `protected function <function name>(ActivityCallbackEvent event)` in `RibActivity` through which child `Activities` can send `ActivityCallbackEvent`

## Context (Environment)

In case we have a logical way of sending an `ActivityCallbackEvent` through any of the system `overriden` methods lets say `onActivityResult` then we have to create a local instance of `callbacksRelay` and `override` the `callbacks()` function to return our local instance. In this case we need to re-implement all the other system notifications as well like `onTrimMemory`, `onLowMemory`. It's not very obvious and can be future error prone

PR #441, opened 2021-04-29
ISSUE-437: Add protected method to post events to callback relay

hamza1592:
- Add method onActivityCallbackEvent
- Change instances of callbacksRelay.accept to use the onActivityCallbackEvent() function

**Description**:
I have described the issue [here](https://github.com/uber/RIBs/issues/437). Briefly we need to provide a method to be able to post events to the `callbacksRelay` so we can have a custom logic behind some callbacks like `onActivityResult`

**Related issue(s)**:
https://github.com/uber/RIBs/issues/437


tyvsmith:
@jbarr21 I think you were involved in offline conversation to close out if we wanted this public API or not. Can you please update here and close out if necessary?

jbarr21:
spoke offline and provided a few potential workarounds to the issue that was being seen

PR #466, opened 2021-07-09
Remove japicmp

jbarr21:
No longer needed since migration to Kotlin has been completed #416


idanakav (approved):

tyvsmith (changes requested):
Wouldn't we want to keep this long term as a useful tool in tracking API changes?

jbarr21:
Ending up keeping this even though the Kotlin migration since it can still help pointing out breaking changes

PR #475, opened 2021-09-03
Add NEW_TASK_REPLACE RouterNavigator flag

hansler:
This flag is similar to the NEW_TASK flag, but it always detaches all existing routers and attaches the new one.

Currently, the DEFAULT flag is the only available flag that detaches the top router if `newStateIsTop` is true. `NEW_TASK_REPLACE` combines this property of `DEFAULT` with the behavior of `NEW_TASK`.

This allows pushing a state with the guarantee that all routers in the stack will be cleared, and that the existing router will be detached.

### DIFF
--- a/android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
+++ b/android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
@@ -59,10 +59,16 @@ interface RouterNavigator<StateT : RouterNavigatorState> {
     NEW_TASK,
 
     /**
+     * Clears the previous stack (no back stack) and pushes the state on to the top of the stack.
+     * If the state is already on the top of the stack, it is detached and replaced.
+     */
+    NEW_TASK_REPLACE,
+
+    /**
      * Remove the top state in the stack if it is not empty and then push a new state to the top of
      * the stack.
      */
-    REPLACE_TOP
+    REPLACE_TOP,
   }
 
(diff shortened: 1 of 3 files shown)


android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
```diff
@@ -58,11 +58,17 @@ interface RouterNavigator<StateT : RouterNavigatorState> {
...
+     * If the state is already on the top of the stack, it is detached and replaced.
+     */
+    NEW_TASK_REPLACE,
+
     /**
      * Remove the top state in the stack if it is not empty and then push a new state to the top of
      * the stack.
      */
-    REPLACE_TOP
+    REPLACE_TOP,
```
jbarr21, line 71: Does clear top plus new task give you the behavior that you're looking for?

hansler: Clear top + new task won’t detach the currently attached RIB if the state name of that state matches the state name of the state being pushed, but NEW_TASK_REPLACE would.

The scenario we’d want to use this for is: We want to be able to navigate a user from one RIB to another. The state name is the same for both the RIBs, but they have different _contexts_ (dynamic dependencies), so we want to detach the existing RIB and attach the new one in order for those different dynamic dependencies to be respected.

Similar to an Activity's on handle new intent - but there's no equivalent for RIBs.

jbarr21 (approved):

PR #480, opened 2021-09-28
TicTacToe sample app in RIBs using Compose

jbarr21:
A sample TicTacToe app built with RIBs and Compose


android/demos/compose/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
...
+    package="com.uber.rib.compose">
+
+    <application
+        android:allowBackup="true"
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:name=".ComposeApplication"
+        android:roundIcon="@mipmap/ic_launcher_round"
+        android:supportsRtl="true"
+        android:theme="@style/Theme.Android">
```
sanchezz93, line 12: should we create a theme for this app?

android/demos/compose/src/main/kotlin/com/uber/rib/compose/ComposeApplication.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.facebook.soloader.SoLoader
+import com.uber.rib.flipper.RibTreePlugin
+
+class ComposeApplication : Application() {
+
+  override fun onCreate() {
+    super.onCreate()
+    SoLoader.init(this, false)
+
+    if (BuildConfig.DEBUG && FlipperUtils.shouldEnableFlipper(this)) {
```
sanchezz93, line 33: Is there an instance where this shouldn't be debug?

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/MainView.kt
```diff
@@ -0,0 +1,76 @@
...
+@Composable
+fun MainView(childContent: MainRouter.ChildContent) {
+  Column(
+    horizontalAlignment = Alignment.CenterHorizontally,
+    verticalArrangement = Arrangement.Top,
+    modifier = Modifier
+      .fillMaxSize()
+      .padding(all = 4.dp)
+      .padding(top = 14.dp)
+      .background(Color(0xFFFFA500))
```
sanchezz93, line 45: Nit : add constants in a const val

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/tic_tac_toe/Board.kt
```diff
@@ -0,0 +1,64 @@
...
+    return (
+      cells[currentRow][0] == theSeed && cells[currentRow][1] == theSeed && cells[currentRow][2] == theSeed ||
+        cells[0][currentCol] == theSeed && cells[1][currentCol] == theSeed && cells[2][currentCol] == theSeed ||
+        currentRow == currentCol && cells[0][0] == theSeed && cells[1][1] == theSeed && cells[2][2] == theSeed ||
+        currentRow + currentCol == 2 && cells[0][2] == theSeed && cells[1][1] == theSeed && cells[2][0] == theSeed
+      )
+  }
+
+  enum class MarkerType {
+    CROSS, NOUGHT
```
sanchezz93, line 48: According to https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md 
enum classes should be PascalCase

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/tic_tac_toe/Board.kt
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  companion object {
+    const val ROWS = 3
+    const val COLS = 3
```
sanchezz93, line 53: Same comment as above, for the companion object

android/demos/compose/src/main/kotlin/com/uber/rib/compose/util/AnalyticsClient.kt
```diff
@@ -0,0 +1,48 @@
...
+    track(id, EventType.IMPRESSION)
+  }
+
+  private fun track(id: String, type: EventType) {
+    val message = "$type for $id @ ${System.currentTimeMillis()}"
+    Toast.makeText(application.applicationContext, message, Toast.LENGTH_SHORT).show()
+    Log.d(this::class.java.simpleName, message)
+  }
+
+  enum class EventType { CLICK, IMPRESSION }
```
sanchezz93, line 37: Same comment as above, change this to PascalCase

android/demos/compose/src/main/AndroidManifest.xml
jbarr21: no, i left it out to keep the focus on the arch

android/demos/compose/src/main/kotlin/com/uber/rib/compose/ComposeApplication.kt
jbarr21: no its just a debug build for devs to try out

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/tic_tac_toe/Board.kt
jbarr21: this is more data model than compose, so i used kotlin naming principles

jbarr21: > Library and app code with a strong existing investment in CAPITALS_AND_UNDERSCORES in their codebase MAY opt for local consistency with that pattern instead.

i opted for local consistency since the objects were closer to business model than compose ui

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/MainView.kt
jbarr21: colors and dimens would typically be in their own files, but since the focus is on arch, i wanted to minimize the jumping around

sanchezz93 (approved):

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/MainInteractor.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class MainInteractor(
+  presenter: ComposePresenter,
+  private val authStream: AuthStream,
+  private val childContent: MainRouter.ChildContent
+) : BasicInteractor<ComposePresenter, MainRouter>(presenter) {
+
+  override fun didBecomeActive(savedInstanceState: Bundle?) {
+    super.didBecomeActive(savedInstanceState)
+    router.view.setContent { MainView(childContent = childContent) }
```
Kamerton12, line 33: Why are we duplicating compose view create here?
We already have done it in local `ComposePresenter`.

Can't we just use `router.view.setContent { presenter.composable() }`?

PR #483, opened 2021-10-28
In compose sample: let child ribs populate their own views, taking the job away from parent router

Ericliu001:
**Description**:
Code organization optimization: making attaching/detaching view the job of the child router and move the logic up to `BasicComposeRouter`. 
Parent Routers do not operate directly on children's views anymore. 

**Related issue(s)**:

### DIFF
--- a/android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/MainRouter.kt
+++ b/android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/MainRouter.kt
@@ -15,10 +15,10 @@
  */
 package com.uber.rib.compose.root.main
 
+import android.view.ViewGroup
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
+import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.ComposeView
 import com.uber.rib.compose.root.main.logged_in.LoggedInRouter
 import com.uber.rib.compose.root.main.logged_out.LoggedOutRouter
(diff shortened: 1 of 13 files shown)


android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/MainRouter.kt
```diff
@@ -27,48 +27,56 @@ import com.uber.rib.core.BasicViewRouter
...
     loggedInRouter?.let {
-      childContent.fullScreenContent = null
       detachChild(it)
     }
     loggedInRouter = null
   }
 
   class ChildContent {
-    internal var fullScreenContent: (@Composable () -> Unit)? by mutableStateOf(null)
+    var fullScreenSlot: MutableState<(@Composable () -> Unit)> = mutableStateOf({})
```
jbarr21, line 80: since our slots are non-nullable, can you please revert back to the delegate usage? (no MutableState wrapper)

jbarr21 (changes requested):

android/libraries/rib-android-compose/src/main/kotlin/com/uber/rib/core/BasicComposeRouter.kt
```diff
@@ -15,7 +15,21 @@
...
 
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+
 open class BasicComposeRouter<I : BasicInteractor<*, *>>(
   val presenter: ComposePresenter,
-  interactor: I
-) : BasicRouter<I>(interactor)
+  interactor: I,
+  val slot: MutableState<(@Composable () -> Unit)>
```
jbarr21, line 24: same here, can remove mutablestate and use property delegate instead of using .value

Ericliu001: synced offline, motif treats type `@Composable () -> Unit` as `unitFunction0()`, as a result, we are not able to pass `@Composable () -> Unit` in parameters as a type.
`
/RIBs/android/demos/compose/build/generated/source/kaptKotlin/debug/com/uber/rib/compose/root/main/logged_in/LoggedInScopeImpl.kt: (85, 85): Type inference failed. Expected type mismatch: inferred type is () -> Unit but @Composable () -> Unit was expected
`

jbarr21 (approved):

PR #487, opened 2021-11-12
fix carthage missing ComponentizedBuilder, MultiStageComponentizedBuilder

ElonPark:
**Description**:
When using Carthage to import RIBs, the two builders below are not available.

1. `ComponentizedBuilder`,
2. `MultiStageComponentizedBuilder`

**Related issue(s)**:
None


ElonPark:
@kovpas Can you please review?

Issue #460, opened 2021-06-25
[Android] Add Coroutines Support


tyvsmith:
Add coroutine APIs to provide lifecycle based scopes and wrap RX apis for consumption with ktx artifact.

kaushalyap:
@tyvsmith side question : Does RIBs use RxJava internally?

tyvsmith:
@kaushalyap Sorry for the late response (missed your question), but yes, it's used internally. I may eventually swap that out, but for now the coroutine support in https://github.com/uber/RIBs/pull/498 is extension based only.

PR #498, opened 2022-04-19
Add initial Coroutines support through dedicated artifact of extension functions

tyvsmith:
This adds APIs to support Coroutines with Ribs.

- This add coroutines support to ScopeProvider based objects (interactors, workers) via extension functions on Autodispose ScopeProvider.
- We add RibDispatcher as a holding place to adapt Rx Schedulers or test schedulers
- A test artifact contains the Junit Rule to setup dispatchers as well as inject TestDispatchers and TestCoroutineScopes into ScopeProvider objects.

Fixes https://github.com/uber/RIBs/issues/460


android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibDispatchers.kt
```diff
@@ -0,0 +1,43 @@
+package com.uber.rib.core
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+object RibDispatchers {
```
tyvsmith, line 6: Looking for feedback on this delegate API.

Will be used to set Dispatchers at app startup like RxJavaPlugins.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutines.kt
```diff
@@ -0,0 +1,20 @@
+package com.uber.rib.core
+
+import com.uber.autodispose.ScopeProvider
+import com.uber.autodispose.coroutinesinterop.asCoroutineScope
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+
+val ScopeProvider.defaultScope: CoroutineScope
```
tyvsmith, line 9: Looking for feedback on 1 scope per default dispatcher as extension functions to interactor.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/CoroutineInteractor.kt
```diff
@@ -0,0 +1,44 @@
...
+  @JvmField protected override var presenter: P
+) : BasicInteractor<P, R>(presenter) {
+
+    /**
+     * Called when attached in main CoroutineScope. The presenter will automatically be added when this happens.
+     *
+     * @param savedInstanceState the saved [Bundle].
+     * @param mainScope the [CoroutineScope] that launched in [RibDispatchers.Main].
+     */
+    abstract suspend fun didBecomeActive(savedInstanceState: Bundle?, mainScope : CoroutineScope)
```
tyvsmith, line 36: Feedback on API name?

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/CoroutineInteractor.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    final override fun didBecomeActive(savedInstanceState: Bundle?) {
```
tyvsmith, line 38: Any reason a developer should be able to override this? Final seems to make sense.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_out/LoggedOutView.kt
```diff
@@ -24,34 +24,34 @@ import androidx.compose.material.Button
...
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.uber.rib.compose.util.EventStream
+import kotlinx.coroutines.launch
 
 @Composable
 fun LoggedOutView(viewModel: State<LoggedOutViewModel>, eventStream: EventStream<LoggedOutEvent>) {
+
+  val coroutineScope = rememberCoroutineScope()
```
tyvsmith, line 39: @jbarr21 any recommended way to call suspend funcs in Compose besides this?

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/CoroutineInteractor.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import kotlinx.coroutines.*
+
+/**
+ * [Interactor] Interactor that launches CoroutineScope on main by default to allow simpler access of suspending functions.
+ *
+ * @param <P> the type of [Presenter].
+ * @param <R> the type of [Router].
+ */
+abstract class CoroutineInteractor<P : Any, R : Router<*>> protected constructor(
```
idanakav, line 26: I wonder if instead of introducing a new class we can use an extension property for the scope.
The main functionality we need is to expose a coroutine scope bounded to the lifecycle of the interactor.

The challenge can be to cache the same scope and reuse it, this will require a modification to the base Interactor class but might be worth exploring.

jbarr21: +1 to no new class

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/CoroutineInteractor.kt
jbarr21: name is fine, but not sure if this is a pattern that we'll want to follow (esp at the cost of req'ing a new base class)

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_out/LoggedOutView.kt
jbarr21: nope, this is all. remember to hold onto it over the course of a composition and launch from it

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
```diff
@@ -0,0 +1,61 @@
...
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.test.TestCoroutineScope
+import kotlin.coroutines.CoroutineContext
+
+@ExperimentalCoroutinesApi
+/**
+ * Allows overriding [ScopeProvider.coroutineScope] with a [TestCoroutineScope] for testing
+ */
+val ScopeProvider.testCoroutineScopeOverride: TestCoroutineScope?
```
tyvsmith, line 32: @jbarr21 @idanakav This is the test utils class for the Scopes. There were multiple use-cases where having the scope be overridden would make since, and it would be harder to test a lifecycle based scope with your interactor otherwise. So I've added this along with some test example code.

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
```diff
@@ -0,0 +1,61 @@
...
+fun ScopeProvider.enableTestCoroutineScopeOverride(context: CoroutineContext = SupervisorJob()) = synchronized(LazyCoroutineScope.values) {
+  LazyCoroutineScope.values[this] = asTestCoroutineScope(context)
+}
+
+fun ScopeProvider.disableTestCoroutineScopeOverride() = synchronized(LazyCoroutineScope.values) {
+  LazyCoroutineScope.values.remove(this)
+}
+
+@ExperimentalCoroutinesApi
+fun ScopeProvider.asTestCoroutineScope(context: CoroutineContext = SupervisorJob()): TestCoroutineScope {
```
tyvsmith, line 49: @ZacSweers @ShaishavGandhi Would you want these asTestCouritineScope function extensions in autodispose as a coroutines test artifact?

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibDispatchers.kt
```diff
@@ -0,0 +1,53 @@
...
+data class TestRibDispatchers(
+  override val Default: TestCoroutineDispatcher = TestCoroutineDispatcher(),
+  override val IO: TestCoroutineDispatcher = TestCoroutineDispatcher(),
+  override val Unconfined: TestCoroutineDispatcher = TestCoroutineDispatcher(),
+  val MainTestDelegate: TestCoroutineDispatcher = TestCoroutineDispatcher()
+) : RibDispatchersProvider {
+
+  fun installTestDispatchers() {
+    // MainTestCoroutineDispatcher is Internal, so we need to wrap it through the main API
+    Dispatchers.setMain(MainTestDelegate)
```
tyvsmith, line 35: This is gross, but has to be done.

android/libraries/rib-coroutines-test/src/test/kotlin/com/uber/rib/core/RibDispatchersTest.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+package com.uber.rib.core
+
+import com.google.common.truth.Truth.assertThat
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import org.junit.Test
+
+@OptIn(ExperimentalCoroutinesApi::class)
```
tyvsmith, line 24: The tests exists in the test artifact class to utilize the Junit rule without having to get too hacky.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.uber.autodispose.ScopeProvider
+import com.uber.autodispose.coroutinesinterop.asCoroutineScope
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import java.util.WeakHashMap
+import kotlin.coroutines.CoroutineContext
+import kotlin.reflect.KProperty
+
+val ScopeProvider.coroutineScope: CoroutineScope by LazyCoroutineScope {
```
jbarr21, line 27: coroutineScope is pretty generic. would it be better to use something more descriptive like `ribScope` like viewmodel uses w/ [viewModelScope](https://developer.android.com/topic/libraries/architecture/coroutines#viewmodelscope)? this would help readability to make it more clear what the scope was

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/LoggedInInteractor.kt
```diff
@@ -38,20 +39,24 @@ class LoggedInInteractor(
...
+          is LoggedInEvent.LogOutClick -> authStream.accept(AuthInfo(false))
+        }
+      }
+      .launchIn(coroutineScope)
 
     router.attachOffGame(authInfo)
   }
 
-  override fun onStartGame() {
+  override suspend fun onStartGame() {
```
jbarr21, line 54: do onStartGame and onGameWon need to be suspending?

android/demos/compose/src/main/kotlin/com/uber/rib/compose/util/EventStream.kt
```diff
@@ -15,13 +15,18 @@
...
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.withContext
 
 class EventStream<T> {
-  private val eventRelay = PublishRelay.create<T>()
+  private val sharedFlow = MutableSharedFlow<T>()
 
-  fun notify(event: T) = eventRelay.accept(event)
+  suspend fun notify(event: T) = withContext(RibDispatchers.Main) {
```
jbarr21, line 27: how come this uses Main but StateStream's dispatch method doesn't. it'd be really nice to have method calls to notify not have to launch with a coroutinescope similar to StateStream.dispatch()

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/tic_tac_toe/TicTacToeInteractor.kt
```diff
@@ -36,58 +37,55 @@ class TicTacToeInteractor(
...
+            )
+          }
+          TicTacToeEvent.XpButtonClick -> TODO("Go somewhere")
+        }
+      }.launchIn(coroutineScope)
   }
 
   interface Listener {
-    fun onGameWon(winnerName: String?)
+    suspend fun onGameWon(winnerName: String?)
```
jbarr21, line 89: why does this need to be suspend?

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
tyvsmith: Yeah, this is something I've gone back and forth on. It's a scope for anything that implements ScopeProvider, so technically not limited to an interactor or to Ribs.

Here's my thinking:
1. The primary autodispose method is ScopeProvider.asCoroutineScope, so the default one as a property would be ScopeProvider.coroutineScope.
2. I would have preferred ScopeProvider.scope, but it would confuse with Motif.
3. If it was ribScope, then it could be confusing if you had worker.ribScope and interactor.ribScope in the same Rib, as they would imply they are the same, but they aren't the same object.

I also debated if these functions should live in this repo at all, vs in Autodispose since they are ScopeProvider extension functions, but in the end all of this work provides a number of util functions we think will work well with Ribs. We may long term remove the autodispose dependency if we decide to implement native coroutine and lifecycle support in Ribs-base.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/LoggedInInteractor.kt
tyvsmith: This was a holdover from a previous experiment with the interactor baseclass. We can remove this now as long as we launch a coroutine in the impl to emit to the sharedflow.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/tic_tac_toe/TicTacToeInteractor.kt
tyvsmith: This was a holdover from a previous experiment with the interactor baseclass. We can remove this now as long as we launch a coroutine in the impl to emit to the sharedflow.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/util/EventStream.kt
tyvsmith: It doesn't and will be removed so much StateStream.

 However, unless we construct with eventStreams with a coroutineScope and they launch them internally (not recommended since we pass these outside of the parent's lifecycle) then they need to be suspending, which means they need to be accessed from a caller using a coroutine. Remember that coroutines are cheap, so it's not a problem like if we had to do it with threads.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/CoroutineInteractor.kt
tyvsmith: Have removed base class.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutines.kt
tyvsmith: Discussed offline, using one scope since there's one lifecycle and the caller can specify other dispatchers in launch command.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/LoggedInInteractor.kt
```diff
@@ -38,10 +40,14 @@ class LoggedInInteractor(
 
   override fun didBecomeActive(savedInstanceState: Bundle?) {
     super.didBecomeActive(savedInstanceState)
+
     eventStream.observe()
-      .ofType<LoggedInEvent.LogOutClick>()
-      .autoDispose(this)
-      .subscribe { authStream.accept(AuthInfo(false)) }
+      .onEach {
```
patrickpsteiger, line 45: Why not use the `filterIsInstance` operator instead of checking the instance in `onEach`?

```kotlin
eventStream.observe()
    .filterIsInstance<LoggedInEvent.LogOutClick>()
    .onEach { authStream.accept(AuthInfo(false)) }
    .launchIn(coroutineScope)
```

jbarr21 (approved):

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/ScoreStream.kt
```diff
@@ -15,27 +15,30 @@
...
+  private val scoresFlow = MutableStateFlow(
     mapOf(
       playerOne to 0,
       playerTwo to 0
     )
   )
 
-  fun addVictory(userName: String) {
-    val scores = (scoresRelay.value ?: emptyMap()).toMutableMap()
+  suspend fun addVictory(userName: String) = withContext(RibDispatchers.IO) {
```
patrickpsteiger, line 33: I think `addVictory` should update the flow atomically, as there's potential to have concurrent coroutines calling `addVictory`.

```kotlin
  suspend fun addVictory(userName: String) = withContext(RibDispatchers.IO) {
    scoresFlow.update { scores ->
      scores.toMutableMap().apply {
        if (userName in scores) {
          set(userName, getValue(userName) + 1)
        }
      }.toMap()
    }
  }
```

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/ScoreStream.kt
```diff
@@ -15,27 +15,30 @@
...
     if (userName in scores) {
       scores[userName] = scores[userName]!! + 1
     }
-    scoresRelay.accept(scores)
+    scoresFlow.emit(scores)
   }
 
-  fun scores(): Observable<Map<String, Int>> {
-    return scoresRelay.hide()
+  fun scores(): Flow<Map<String, Int>> {
```
patrickpsteiger, line 41: Why not expose the read-only `StateFlow` instead? Also, instead of wrapping the `MutableStateFlow` at every call on `scores()`, we could have a backing property instead.

```kotlin
  private val _scoresFlow = MutableStateFlow(
    mapOf(
      playerOne to 0,
      playerTwo to 0
    )
  )
  private val scoresFlow = _scoresFlow.asStateFlow()

  // fun addVictory ...
  
  fun scores() = scoresFlow
```

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/off_game/OffGameInteractor.kt
```diff
@@ -35,28 +35,29 @@ class OffGameInteractor(
...
   override fun didBecomeActive(savedInstanceState: Bundle?) {
     super.didBecomeActive(savedInstanceState)
+
     eventStream.observe()
-      .ofType<OffGameEvent.StartGame>()
-      .autoDispose(this)
-      .subscribe {
-        listener.onStartGame()
-      }
+      .onEach {
```
patrickpsteiger, line 40: Should use `filterIsInstance` ?

```kotlin
eventStream.observe()
    .filterIsInstance<OffGameEvent.StartGame>()
    .onEach { listener.onStartGame() }
    .launchIn(coroutineScope)
```

android/demos/compose/src/main/kotlin/com/uber/rib/compose/util/StateStream.kt
```diff
@@ -14,16 +14,16 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asSharedFlow
 
 class StateStream<T : Any>(default: T) {
-  private val stateRelay = BehaviorRelay.createDefault(default)
+  private val stateFlow = MutableStateFlow(default)
 
-  fun dispatch(viewModel: T) = stateRelay.accept(viewModel)
+  suspend fun dispatch(viewModel: T) = stateFlow.emit(viewModel)
```
patrickpsteiger, line 24: Should this be non-suspending and use `stateFlow.update { viewModel }` ?

android/demos/compose/src/main/kotlin/com/uber/rib/compose/util/StateStream.kt
```diff
@@ -14,16 +14,16 @@
...
 
-  fun observe(): Observable<T> = stateRelay.hide()
+  fun observe(): Flow<T> = stateFlow.asSharedFlow()
```
patrickpsteiger, line 26: Why not expose it as a read-only `StateFlow` ?

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/LoggedInInteractor.kt
tyvsmith: The original approach had this, however with the when() on sealed classes we can compiler guarantee all use-cases are covered so would be preferred.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/ScoreStream.kt
tyvsmith: That's a reasonable suggestion

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/ScoreStream.kt
tyvsmith: SharedFlow was typo. Fixed. I like the idea of the backing field, I'll move all of the Streams to that.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/off_game/OffGameInteractor.kt
tyvsmith: See above comment on sealed classes.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/main/logged_in/LoggedInInteractor.kt
patrickpsteiger: Understood -- in cases where multiple children of sealed classes are dealt with, we'd just add new cases to `when`. You can go ahead and resolve all comments related to `filterIsInstance`.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  RibCoroutinesConfig.exceptionHandler?.let {
+    context += RibCoroutinesConfig.exceptionHandler!!
+  }
+  asCoroutineScope(context)
+}
+
+internal class LazyCoroutineScope(val initializer: ScopeProvider.() -> CoroutineScope) {
+  companion object {
+    val values = WeakHashMap<ScopeProvider, CoroutineScope>()
```
patrickpsteiger, line 45: I understand the map is exposed just for internal tests, but instead of exposing the whole map, I'd prefer exposing a limited API with getter and setter for a specific scope. This would prohibit insane operations such as `LazyCoroutineScope.values.clear()`

```kotlin
private val values = WeakHashMap<ScopeProvider, CoroutineScope>()

// Test APIs
internal operator fun get(provider: ScopeProvider) = values[provider]
internal operator fun set(provider: ScopeProvider, scope: CoroutineScope) { 
  values[provider] = scope
}
```

Usage:

```kotlin
fun exampleUsage(scopeProvider: ScopeProvider, newCoroutineScope: CoroutineScope) {
  val current = LazyCoroutineScope[scopeProvider]
  LazyCoroutineScope[scopeProvider] = newCoroutineScope
}
```

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibDispatchers.kt
```diff
@@ -0,0 +1,147 @@
...
+import kotlin.coroutines.ContinuationInterceptor
+
+object RibDispatchers : RibDispatchersProvider {
+  override val Default: CoroutineDispatcher get() = RibCoroutinesConfig.dispatchers.Default
+  override val Main: MainCoroutineDispatcher get() = RibCoroutinesConfig.dispatchers.Main
+  override val IO: CoroutineDispatcher get() = RibCoroutinesConfig.dispatchers.IO
+  override val Unconfined: CoroutineDispatcher get() = RibCoroutinesConfig.dispatchers.Unconfined
+}
+
+data class DefaultRibDispatchers(
```
patrickpsteiger, line 34: Wondering if we do want this to be a `data class` -- why not a regular class instead?

android/libraries/rib-coroutines/README.md
```diff
@@ -0,0 +1,37 @@
...
+```kotlin
+RibCoroutinesConfig.dispatchers = DefaultRibDispatchers()
+
+RibCoroutinesConfig.exceptionHandler = CoroutineExceptionHandler { _, exception ->
+    Log.e(exception)
+}
+```
+
+
+Lifecycle based components such as Interactors and Workers receive a dedicated coroutineScope for usage.
```
patrickpsteiger, line 26: That's a good opportunity to clarify that the `coroutineScope` is cancelled when the `ScopeProvider` completes.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
```diff
@@ -0,0 +1,50 @@
...
+ * This scope will be canceled when ScopeProvider is completed
+ *
+ * This scope is bound to
+ * [RibDispatchers.Main.immediate][kotlinx.coroutines.MainCoroutineDispatcher.immediate]
+ */
+val ScopeProvider.coroutineScope: CoroutineScope by LazyCoroutineScope {
+  var context: CoroutineContext = SupervisorJob() + RibDispatchers.Main.immediate + CoroutineName("${this::class.simpleName}:coroutineScope")
+
+  RibCoroutinesConfig.exceptionHandler?.let {
+    context += RibCoroutinesConfig.exceptionHandler!!
```
bhargavmaniyar, line 38: super nit:-  can we replace !! with "context += it "?

android/libraries/rib-coroutines/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+
+apply plugin: "org.jetbrains.kotlin.jvm"
+apply plugin: "com.vanniktech.maven.publish"
+
+sourceCompatibility = deps.build.javaVersion.toString()
+targetCompatibility = deps.build.javaVersion.toString()
+
```
patrickpsteiger, line 23: This is personal preference, but for libraries, I think having explicit visibility modifiers is better -- we don't want to accidentally expose API.

```kotlin
kotlin {
    explicitApi()
}
```

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibDispatchers.kt
tyvsmith: Not strongly opinionated on this one.

Since it's being used to hold dispatchers as an object bag, the equality check *may* be useful by consumers in more complex cases of adapting during testing scenarios or during initialization of an transforming layer from Rx schedulers to Dispatchers.

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
tyvsmith: Since it's all internal, a consumer could never call it. Never the less, this is a sane recommendation that I'll update to.

idanakav (approved):

Issue #502, opened 2022-04-24
[Intellij Plugin] Broken JNA on Arm 64 JDK


tyvsmith:
The ribs Intellij plugins won't built on arm64 machines. Needs at least JNA updated, potentially other investigation as well.

Stacktrace:
`[gradle-intellij-plugin :rib-intellij-plugin:buildSearchableOptions] Cannot resolve builtin JBR 11_0_3b304.10. Falling local Java.
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.intellij.ide.ClassUtilCore to field sun.net.www.protocol.jar.JarFileFactory.fileCache
WARNING: Please consider reporting this to the maintainers of com.intellij.ide.ClassUtilCore
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
2022-04-24 13:07:42,866 [     18]   WARN -        #com.intellij.idea.Main - Unable to load JNA library (OS: Mac OS X 12.3.1)
java.lang.UnsatisfiedLinkError: Can't load library: /Users/tys/.gradle/wrapper/dists/gradle-7.0.2-all/7era6s5ay7zsbhuvl0oc9g94s/gradle-7.0.2/caches/intellij/system/tmp/jna1075360584433724666.tmp
        at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2633)
        at java.base/java.lang.Runtime.load0(Runtime.java:768)
        at java.base/java.lang.System.load(System.java:1837)
        at com.sun.jna.Native.loadNativeDispatchLibraryFromClasspath(Native.java:1018)
        at com.sun.jna.Native.loadNativeDispatchLibrary(Native.java:988)
        at com.sun.jna.Native.<clinit>(Native.java:195)
        at com.intellij.jna.JnaLoader.load(JnaLoader.java:16)
        at com.intellij.idea.StartupUtil.loadSystemLibraries(StartupUtil.java:543)
        at com.intellij.idea.StartupUtil.lambda$prepareAndStart$4(StartupUtil.java:205)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #504, opened 2022-04-24
Makes Intellij plugin compile on latest [WIP]

tyvsmith:
Can now build, compile, and install the intellij plugin now (cannot on main), however it appears the adb bridge itself no longer works and needs to be updated.

Screen Shot 2022-04-24 at 3 15 41 PM


tyvsmith:
This should resolve #502 once it lands. We need to get the functionality fixed first though.

tyvsmith:
Abandoning in favor of #508 which already landed.

PR #509, opened 2022-05-20
Fix unexpected dependency version of CocoaPods

chorim:
**Description**:
I'm found a bug while installing RIBs using CocoaPods.

I have confirmed that the results are different from when I used a different package manager(such as SwiftPM, Carthage).

All installation methods are looking at the branch "main".

**SwiftPM**:
image

**Carthage**:
image

**CocoaPods**:
image

This issue occurs because the "CocoaPods" dependency requirement version notation is different.
https://guides.cocoapods.org/syntax/podspec.html#dependency

SwiftPM:
https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#package-dependency

Carthage:
https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#cartfile

I think CocoaPods are a bug, so I open a Pull Request.

**Related issue(s)**:


rudro:
Have landed https://github.com/uber/RIBs/pull/534 which should address

PR #520, opened 2022-08-03
Migrate com.nhaarman.mockitokotlin2 to org.mockito.kotlin

SergeySmykovskyi:
**Description**:  mockito-kotlin was created and developed by [nhaarman@](https://github.com/nhaarman) after which the repository was integrated into the official Mockito GitHub organization.

### DIFF
--- a/android/gradle/dependencies.gradle
+++ b/android/gradle/dependencies.gradle
@@ -99,7 +99,7 @@ def androidx = [
 def test = [
     junit: "junit:junit:4.12",
     mockito: "org.mockito:mockito-core:2.27.0",
-    mockitoKotlin: "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0",
+    mockitoKotlin: "org.mockito.kotlin:mockito-kotlin:4.0.0",
     compileTesting: "com.google.testing.compile:compile-testing:0.17",
     truth: "com.google.truth:truth:0.43",
 ]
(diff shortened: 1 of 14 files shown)


android/gradle/dependencies.gradle
```diff
@@ -98,8 +98,8 @@ def androidx = [
 
 def test = [
     junit: "junit:junit:4.12",
-    mockito: "org.mockito:mockito-core:2.27.0",
-    mockitoKotlin: "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0",
+    mockito: "org.mockito:mockito-core:4.6.1",
+    mockitoKotlin: "org.mockito.kotlin:mockito-kotlin:4.0.0",
```
tyvsmith, line 102: Do we need to keep these two versions in lockstep?

rafaeltoledo: No, mockito-kotlin do not follow the main mockito artifacts ([link](https://github.com/mockito/mockito-kotlin/tags))

SergeySmykovskyi: Currently `mockito-kotlin` based on Mockito 3.12.4. Starting from Mockito 4+, methods which were deprecated in 3+ was removed, so better to also bump-up `mockito-core` for using latest API in codebase. It will make future bump-up easier.

jbarr21 (approved):

PR #522, opened 2022-09-19
add ios-objc files

sevli:
**Description**:
Hi, I am a iOS Developer from DiDi Global. Because most iOS developers use Objctive-C for development, so we want to contribute Objctive-C code. 

Look forward to your soon reply!

**Related issue(s)**:


alanzeino:
Hi Vincent!

Thank you for your contribution.

I want to start off by saying this is AMAZING and I can't believe someone took the time to port this to Objective-C. Well done! Honestly it is so impressive.

That being said, our belief right now is that we likely would not add this support to RIBs today, since Uber is mostly Swift, so we could not adequately test this code on our side. By adding it to the RIBs repo we would implicitly be advertising this code's 'robustness' so to say, and we can't really do that ourselves since we don't have any meaningful way to use this.

However, I think it's totally fine for you to publish this as a fork and as a new library on Cocoapods, etc. We're totally happy with you making this available for other developers in the iOS community to use.

I would even be happy to link to your implementation in our README, so that people can find it.

alanzeino:
Going to close this for now, if you want us to link to this let me know @sevli

Gdakgdak22 (approved):

Gdakgdak22 (approved):

PR #525, opened 2022-10-19
Ty/kotlin 1.7 upgrade

tyvsmith:
Upgrades all libraries required for Kotlin language version upgrade:

- Kotlin -> 1.7.10
- Coroutines -> 1.6.4
- Dagger -> 2.43.2
- Compose Compiler -> 1.3.0
- Compose libraries -> 1.2.1
- Compose percent ->1.0.0
- Compose savedState - >1.2.0

Additionally ,we removed some functionality based on Coroutines changing APIs.

- Removes Deprecated TestCoroutineScope and runBlockingTest in favor of TestScope and runTest per https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/MIGRATION.md
- Removes TestRule as functioning TestCoroutineScope (or TestScope). The value of the original intent is low now because A) Run blocking follows structured concurrency hierarchical error reporting properly and B) We can no longer enforce a cleanup in a finishing block of the TestRule, we would instead need to introduce a new wrapping Lambda to replace runTest (The standard API)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/BasicInteractor.kt
```diff
@@ -22,5 +22,5 @@ package com.uber.rib.core
  * @param <R> the type of [Router].
  */
 abstract class BasicInteractor<P : Any, R : Router<*>> protected constructor(
-  @JvmField protected open var presenter: P
+  protected open var presenter: P
```
jbarr21, line 25: Is this still binary compatible? Does it still integrate into the amount of repo without (or limited to a handful) issues?

jbarr21 (approved):

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/BasicInteractor.kt
tyvsmith: It is not binary compatible. However the current bug that was fixed in Kotlin created invalid JVM code that resulted in a runtime crash. This moves the crash to compile time, so as long as build succeeds, it should be fine.

https://youtrack.jetbrains.com/issue/KT-32753

Build and test of major apps against the upgraded version in Android repo was successful.

PR #536, opened 2023-01-17
Clear cached `CoroutineScope` instance once its job completes

### LINKED ISSUE #535: ScopeProvider.coroutineScope does not support detaching-reattaching on ScopeProvider components (RIB)
**Problem Statement** 
There's a behavior of `ScopeProvider.coroutineScope` that I consider to be unexpected and a bug in its caching logic.

The `ScopeProvider.coroutineScope` extension provides a `CoroutineScope` backed by the `Completable` provided by `ScopeProvider.requestScope()`. The scope `Completable` completes when the corresponding detach callback (willResignActive for Interactor) is called, and likewise, the `CoroutineScope` backed by the `Completable` gets canceled.

If the same `ScopeProvider` instance gets attached again, on the attach callbacks (`didBecomeActive`), `ScopeProvider.requestScope()` will return a new `Completable` (expected, remember that the previous `Completable` got completed and there's no "uncompleting" it). That behavior allows users to launch scoped jobs on the attach callback and expect them to be re-launched on the event of detach-reattach. On the other hand, `ScopeProvider.coroutineScope` will not use this new completable: **It is cached to the class instance and will return the old, canceled `CoroutineScope` (unexpected!)**.

We should probably expect that the `ScopeProvider.coroutineScope` extension behaves similarly to `ScopeProvider.requestScope()` in lifecycle terms.

**Concrete example**
On a concrete example, it is unexpected that, e.g., in a `ScopeProvider` component, while a scoped Rx subscription (bound to the `ScopeProvider` by `autoDispose(this)`) will get restarted on the event of detach-reattach (as long as the subscription happens in the attach callback), a Flow collection coroutine launched by `launchIn(coroutineScope)` in the attach callback will not be re-launched on the event of detach-reattach. (Note that a Flow collection was given as an example, but this behavior applies to any coroutine launched in the scope of `ScopeProvider.coroutineScope`).

This unexpected behavior might be the source of hard-to-debug bugs and the API, so I propose we align `ScopeProvider.coroutineScope` behavior to the already established `ScopeProvider.requestScope()` behavior, and this can be achieved by changing the `CoroutineScope` caching logic.

**Proposed fix**
The current code (at the time of writing) lives at `com.uber.rib.core` and is:

```kotlin
public val ScopeProvider.coroutineScope: CoroutineScope by LazyCoroutineScope<ScopeProvider> {
 val context: CoroutineContext = SupervisorJob() +
   RibDispatchers.Main.immediate +
   CoroutineName("${this::class.simpleName}:coroutineScope") +
   (RibCoroutinesConfig.exceptionHandler ?: EmptyCoroutineContext)

 asCoroutineScope(context)
}

internal class LazyCoroutineScope<This : Any>(val initializer: This.() -> CoroutineScope) {
 companion object {
   private val values = WeakHashMap<Any, CoroutineScope>()
   // Used to get and set Test overrides from rib-coroutines-test utils
   operator fun get(provider: Any) = values[provider]
   operator fun set(provider: Any, scope: CoroutineScope?) {
     values[provider] = scope
   }
 }
 operator fun getValue(thisRef: This, property: KProperty<*>): CoroutineScope =
     synchronized(LazyCoroutineScope) {
       return values.getOrPut(thisRef) { thisRef.initializer() }
     }
}
```

I propose synchronously clearing the cache on the event of `CoroutineScope` completion. This can be achieved by installing a completion handler in the `CoroutineScope` created in `LazyCoroutineScope`.

```kotlin
public val ScopeProvider.coroutineScope:
    CoroutineScope by LazyCoroutineScope<ScopeProvider>({ requestScope() }) {
 val context: CoroutineContext = SupervisorJob() +
   RibDispatchers.Main.immediate +
   CoroutineName("${this::class.simpleName}:coroutineScope") +
   (RibCoroutinesConfig.exceptionHandler ?: EmptyCoroutineContext)

 asCoroutineScope(context)
}

internal class LazyCoroutineScope<This : Any>(
    val keyProducer: This.() -> Any = { this },
    val initializer: This.() -> CoroutineScope
) {
 companion object {
   private val values = WeakHashMap<Any, CoroutineScope>()
   // Used to get and set Test overrides from rib-coroutines-test utils
   operator fun get(provider: Any) = values[provider]
   operator fun set(provider: Any, scope: CoroutineScope?) {
     values[provider] = scope
   }
 }
  operator fun getValue(thisRef: This, property: KProperty<*>): CoroutineScope = synchronized(LazyCoroutineScope) {
    return values.getOrPut(thisRef) {
      thisRef.initializer().apply {
        coroutineContext.job.invokeOnCompletion {
          synchronized(LazyCoroutineScope) {
            set(thisRef, null)
          }
        }
      }
    }
  }
}
```

**Previously considered fix**
At first, I proposed a simple fix where we cache the `CoroutineScope` by the `ScopeProvider.requestScope()` `Completable` object instead of caching by the class instance, as below:

```kotlin
public val ScopeProvider.coroutineScope:
    CoroutineScope by LazyCoroutineScope<ScopeProvider>({ requestScope() }) {
 val context: CoroutineContext = SupervisorJob() +
   RibDispatchers.Main.immediate +
   CoroutineName("${this::class.simpleName}:coroutineScope") +
   (RibCoroutinesConfig.exceptionHandler ?: EmptyCoroutineContext)

 asCoroutineScope(context)
}

internal class LazyCoroutineScope<This : Any>(
    val keyProducer: This.() -> Any = { this },
    val initializer: This.() -> CoroutineScope
) {
 companion object {
   private val values = WeakHashMap<Any, CoroutineScope>()
   // Used to get and set Test overrides from rib-coroutines-test utils
   operator fun get(provider: Any) = values[provider]
   operator fun set(provider: Any, scope: CoroutineScope?) {
     values[provider] = scope
   }
 }
 operator fun getValue(thisRef: This, property: KProperty<*>): CoroutineScope = 
     synchronized(LazyCoroutineScope) {
       return values.getOrPut(thisRef.keyProducer()) { thisRef.initializer() }
     }
}
```

The issue with this approach is `ScopeProvider.requestScope()` always returns a new instance of `Completable` every time it is called, effectively making cache useless: a new instance of `CoroutineScope` would always be returned at every call to `ScopeProvider.coroutineScope`.

psteiger:
Fixes #535.

**Description**:
Fixes `ScopeProvider.coroutineScope` lifecycle to better align with the backing `ScopeProvider` lifecycle behavior.

With this fix, a fresh, non-cancelled instance of `CoroutineScope` is delivered by `ScopeProvider.coroutineScope` every time the previously cached instance  is completed/canceled. Every scope instance is synchronously **cleared from cache** once it is completed/canceled. This is implemented by installing a `CompletionHandler` on the `CoroutineScope` `Job` that **clears the cache when scope is completed**.

This new behavior of clearing the cache upon completion effectively allows for re-launching coroutines on the attach callback when an attach-detach-attach cycle occurs on `ScopeProvider` components with a lifecycle.

Re-launching coroutines upon reattachment is not possible in current implementation (without this fix): a single `CoroutineScope` is cached for the whole life of the instance, and `ScopeProvider.coroutineScope` delivers a stale, cancelled instance  of `CoroutineScope` after reattaching the component (even though `ScopeProvider.requestScope()` delivers a new `Completable`.

To be clear, the correct reattachment behavior is already in place for Rx subscriptions that `autoDispose` by `ScopeProvider`:  they are backed by the `Completable`, which work fine on attach-detach-reattach cycles. This PR makes coroutines behavior consistent with the currently established behavior in RIBs.

**Related issue(s)**: #535

### DIFF
--- a/android/libraries/rib-coroutines-test/src/test/kotlin/com/uber/rib/core/RibScopesTest.kt
+++ b/android/libraries/rib-coroutines-test/src/test/kotlin/com/uber/rib/core/RibScopesTest.kt
@@ -147,6 +147,24 @@ internal class RibScopesTest {
     assertThat(realScope2).isNotInstanceOf(TestScope::class.java)
   }
 
+  @Test
+  fun testScopeReattaching() {
+    val interactor = object : BasicInteractor<Presenter, Router<*>>(mock()) {}
+    with(interactor) {
+      dispatchAttach(null)
+      with(coroutineScope) {
+        assertThat(isActive).isTrue()
+        dispatchDetach()
+        // after dispatching detach, we expect the same instance captured before to be inactive
+        assertThat(isActive).isFalse()
+      }
+      dispatchAttach(null)
+      // The previous instance of coroutineScope is permanently cancelled,
+      // but ScopeProvider.coroutineScope should now return a new, active instance.
+      assertThat(coroutineScope.isActive).isTrue()
+    }
+  }
+
   private class TestUncaughtExceptionCaptor : CoroutineExceptionHandler {
     var exceptions = mutableListOf<Throwable>()
 
--- a/android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
+++ b/android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/RibCoroutineScopes.kt
@@ -21,6 +21,7 @@ import com.uber.autodispose.coroutinesinterop.asCoroutineScope
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.job
 import java.util.WeakHashMap
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
@@ -69,6 +70,14 @@ internal class LazyCoroutineScope<This : Any>(val initializer: This.() -> Corout
     }
   }
   operator fun getValue(thisRef: This, property: KProperty<*>): CoroutineScope = synchronized(LazyCoroutineScope) {
-    return values.getOrPut(thisRef) { thisRef.initializer() }
+    return values.getOrPut(thisRef) {
+      thisRef.initializer().apply {
+        coroutineContext.job.invokeOnCompletion {
+          synchronized(LazyCoroutineScope) {
+            values.remove(thisRef)
+          }
+        }
+      }
+    }
   }
 }


tyvsmith:
This looks great, thanks for the contribution! Before landing, can you run through a test against the internal Android repo with app on some coroutine flows.

psteiger:
@tyvsmith responded with manual test results in internal communication tool.

tyvsmith (approved):
Looks good, thanks for the contribution.

PR #541, opened 2023-03-07
Make all `TestDispatcher`s in `TestRibDispatchers` use the same `TestCoroutineScheduler`

### LINKED ISSUE #540: `RibCoroutinesRule` creates one scheduler per dispatcher.
Documentation for `StandardTestDispatcher` and `UnconfinedTestDispatcher` say:

> If no scheduler is passed as an argument, Dispatchers.Main is checked, and if it was mocked with a TestDispatcher via Dispatchers.setMain, the TestDispatcher.scheduler of the mock dispatcher is used; if Dispatchers.Main is not mocked with a TestDispatcher, a new TestCoroutineScheduler is created.

When we are using `Dispatchers.setMain(TestDispatcher)`, we can assume that all dispatchers created *after* the fact will use the same `TestCoroutineScheduler` contained in the `TestDispatcher` set as `Main`.

However, our `TestRibDispatchers` creates the dispatchers *before* setting the main dispatcher with `Dispatchers.setMain`, meaning each one will create its own `TestCoroutineScheduler` (as per docs).

The current implementation for `TestRibDispatchers` makes the following test fail:
```kotlin
@Test
fun testScheduler() = runTest {
  var reached = false
  launch(RibDispatchers.Default) {
    delay(100)
    reached = true
  }
  advanceTimeBy(100) // this will *not* advance the time of `RibDispatchers.Default`, as it has its own scheduler!
  runCurrent()
  assertThat(reached).isTrue()
}
```

I propose the following fix in `TestRibDispatchers`:

```kotlin
public data class TestRibDispatchers(
  // create our testScheduler to be shared by all dispatchers
  private val testScheduler: TestCoroutineScheduler = TestCoroutineScheduler(), 
  // use the testScheduler in all dispatchers created before `Dispatchers.setMain`
  override val Default: TestDispatcher = StandardTestDispatcher(testScheduler),
  override val IO: TestDispatcher = StandardTestDispatcher(testScheduler),
  override val Unconfined: TestDispatcher = UnconfinedTestDispatcher(testScheduler),
  val MainTestDelegate: TestDispatcher = StandardTestDispatcher(testScheduler)
) : RibDispatchersProvider { 
  // ...
}
```

Then, consumers of the API can rest assured every time he's skipping delays, it will apply not only to the main dispatcher, but to all dispatchers in `RibDispatchers`.

psteiger:
Fixes #540.

**Description**: This introduces consistent behavior when advancing the virtual time in a `runTest` block: all dispatchers from `RibDispatchers` will advance in time, as opposed to the current implementation where only the main dispatcher will advance in time.

**Related issue(s)**: #540


tyvsmith (approved):
This seems reasonable.  Can you use a locally built release against the Uber android repo to verify existing compatibility? Once you let me know, we'll merge and cut a release with this fix and the most recent cancellation fix.

tyvsmith:
Talked offline and @psteiger verified within local repo.

Issue #542, opened 2023-03-08
Remove Rxjava from core Ribs artifacts


tyvsmith:
[Placeholder ticket for high level work]
1. Remove RxJava from core artifacts and replace with coroutines
2. Introduce standalone Rxjava2 and RxJava3 libraries that legacy consumers can use.

tyvsmith:
This would also have the benefit of removing the dependency on autodispose for coroutines usage.

PR #544, opened 2023-03-25
Replacing some Behavior/Publish Relay usage in core artifacts with coroutines

tyvsmith:
Beginning internal addressing of Rx usages in core artifacts.

- Currently replacing heavy use of Behavior/Publish Relays on internal code to track lifecycle state.
- Public APIs are still wrapped to observables.

- [x] Verify against internal test suite
- [x] Verify against app core flow functional expectations

https://github.com/uber/RIBs/issues/542

### DIFF
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
@@ -37,6 +37,7 @@ import com.uber.rib.core.lifecycle.ActivityLifecycleEvent
 import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.create
 import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.createOnCreateEvent
 import io.reactivex.CompletableSource
+import io.reactivex.Observable
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.rx2.asObservable
@@ -50,17 +51,17 @@ abstract class RibActivity : CoreAppCompatActivity(), ActivityStarter, Lifecycle
   private val callbacksObservable = callbacksFlow.asObservable()
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun lifecycle() = lifecycleObservable
+  override fun lifecycle(): Observable<ActivityLifecycleEvent> = lifecycleObservable
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun callbacks() = callbacksObservable
+  override fun callbacks(): Observable<ActivityCallbackEvent> = callbacksObservable
 
   override fun correspondingEvents(): CorrespondingEventsFunction<ActivityLifecycleEvent> {
     return ACTIVITY_LIFECYCLE
   }
 
-  override fun peekLifecycle(): ActivityLifecycleEvent {
-    return lifecycleFlow.replayCache.last()
+  override fun peekLifecycle(): ActivityLifecycleEvent? {
+    return lifecycleFlow.replayCache.lastOrNull()
   }
 
   override fun requestScope(): CompletableSource {
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
@@ -37,7 +37,6 @@ import com.uber.rib.core.lifecycle.ActivityLifecycleEvent
 import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.create
 import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.createOnCreateEvent
 import io.reactivex.CompletableSource
-import io.reactivex.Observable
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.rx2.asObservable
@@ -47,16 +46,14 @@ abstract class RibActivity : CoreAppCompatActivity(), ActivityStarter, Lifecycle
   private var router: ViewRouter<*, *>? = null
   private val lifecycleFlow = MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
   private val callbacksFlow = MutableSharedFlow<ActivityCallbackEvent>(0, 1, BufferOverflow.DROP_OLDEST)
+  private val lifecycleObservable = lifecycleFlow.asObservable()
+  private val callbacksObservable = callbacksFlow.asObservable()
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun lifecycle(): Observable<ActivityLifecycleEvent> {
-    return lifecycleFlow.asObservable()
-  }
+  override fun lifecycle() = lifecycleObservable
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun callbacks(): Observable<ActivityCallbackEvent> {
-    return callbacksFlow.asObservable()
-  }
+  override fun callbacks() = callbacksObservable
 
   override fun correspondingEvents(): CorrespondingEventsFunction<ActivityLifecycleEvent> {
     return ACTIVITY_LIFECYCLE
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
@@ -21,9 +21,6 @@ import android.content.res.Configuration
 import android.os.Build
 import android.view.ViewGroup
 import androidx.annotation.CallSuper
-import com.jakewharton.rxrelay2.BehaviorRelay
-import com.jakewharton.rxrelay2.PublishRelay
-import com.jakewharton.rxrelay2.Relay
 import com.uber.autodispose.lifecycle.CorrespondingEventsFunction
 import com.uber.autodispose.lifecycle.LifecycleEndedException
 import com.uber.autodispose.lifecycle.LifecycleScopeProvider
@@ -41,22 +38,24 @@ import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.create
 import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.createOnCreateEvent
 import io.reactivex.CompletableSource
 import io.reactivex.Observable
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.rx2.asObservable
 
 /** Base implementation for all VIP [android.app.Activity]s. */
 abstract class RibActivity : CoreAppCompatActivity(), ActivityStarter, LifecycleScopeProvider<ActivityLifecycleEvent>, RxActivityEvents {
   private var router: ViewRouter<*, *>? = null
-  private val lifecycleBehaviorRelay = BehaviorRelay.create<ActivityLifecycleEvent>()
-  private val lifecycleRelay: Relay<ActivityLifecycleEvent> = lifecycleBehaviorRelay.toSerialized()
-  private val callbacksRelay = PublishRelay.create<ActivityCallbackEvent>().toSerialized()
+  private val lifecycleFlow = MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
+  private val callbacksFlow = MutableSharedFlow<ActivityCallbackEvent>(0, 1, BufferOverflow.DROP_OLDEST)
 
   /** @return an observable of this activity's lifecycle events. */
   override fun lifecycle(): Observable<ActivityLifecycleEvent> {
-    return lifecycleRelay.hide()
+    return lifecycleFlow.asObservable()
   }
 
   /** @return an observable of this activity's lifecycle events. */
   override fun callbacks(): Observable<ActivityCallbackEvent> {
-    return callbacksRelay.hide()
+    return callbacksFlow.asObservable()
   }
 
   override fun correspondingEvents(): CorrespondingEventsFunction<ActivityLifecycleEvent> {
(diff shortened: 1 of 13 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -85,12 +90,14 @@ object WorkerBinder {
...
-    val workerLifecycle = mappedLifecycle
-      .mergeWith(unbindSubject)
-      .takeUntil { workerEvent: WorkerEvent -> workerEvent === WorkerEvent.STOP }
+    val unbindFlow = MutableSharedFlow<WorkerEvent>(0, 1, BufferOverflow.DROP_OLDEST)
+
+    val workerLifecycle = merge(mappedLifecycle.asFlow(), unbindFlow)
+      .transformWhile { emit(it) ; it != WorkerEvent.STOP }
+      .asObservable()
+
     bindToWorkerLifecycle(workerLifecycle, worker)
```
tyvsmith, line 99: I prototyped a copy of bindToWorkerLifecycle that has to launch to match events, and we end up in race condtion since a subscribe on main thread is guaranteed to register before the testing code calls it. Need to look into if we should block for registering the lambda but not on the terminal event.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -41,30 +38,32 @@ import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.create
...
-  private val lifecycleBehaviorRelay = BehaviorRelay.create<ActivityLifecycleEvent>()
-  private val lifecycleRelay: Relay<ActivityLifecycleEvent> = lifecycleBehaviorRelay.toSerialized()
-  private val callbacksRelay = PublishRelay.create<ActivityCallbackEvent>().toSerialized()
+  private val lifecycleFlow = MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
+  private val callbacksFlow = MutableSharedFlow<ActivityCallbackEvent>(0, 1, BufferOverflow.DROP_OLDEST)
 
   /** @return an observable of this activity's lifecycle events. */
   override fun lifecycle(): Observable<ActivityLifecycleEvent> {
-    return lifecycleRelay.hide()
+    return lifecycleFlow.asObservable()
```
psteiger, line 53: Nit, kind of out of scope of the change: I generally prefer to have these cold streams defined as properties instead of doing a new conversion (new allocation!) at every method call, to reduce the number of allocations. Consider:

```kotlin
val flow = MutableSharedFlow<Int>(1, 0, BufferOverflow.DROP_OLDEST)
fun getObservable() = flow.asObservable()

fun main(args: Array<String>) {
    repeat(100) {
        println(getObservable())
    }
}
```
Translates to 1 new allocation every time method is called:
```
io.reactivex.internal.operators.observable.ObservableCreate@7a30d1e6
io.reactivex.internal.operators.observable.ObservableCreate@5891e32e
io.reactivex.internal.operators.observable.ObservableCreate@cb0ed20
io.reactivex.internal.operators.observable.ObservableCreate@8e24743
io.reactivex.internal.operators.observable.ObservableCreate@74a10858
```

if we allocate it as a property once:

```kotlin
val flow = MutableSharedFlow<Int>(1, 0, BufferOverflow.DROP_OLDEST)
val flowAsObservable = flow.asObservable()
fun getObservable() = flowAsObservable

fun main(args: Array<String>) {
    repeat(100) {
        println(getObservable())
    }
}
```
Only one allocation is done:
```
io.reactivex.internal.operators.observable.ObservableCreate@587d1d39
io.reactivex.internal.operators.observable.ObservableCreate@587d1d39
io.reactivex.internal.operators.observable.ObservableCreate@587d1d39
io.reactivex.internal.operators.observable.ObservableCreate@587d1d39
io.reactivex.internal.operators.observable.ObservableCreate@587d1d39
```

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -177,13 +176,15 @@ abstract class RibActivity : CoreAppCompatActivity(), ActivityStarter, Lifecycle
...
+    lifecycleFlow.tryEmit(create(ActivityLifecycleEvent.Type.USER_LEAVING))
     super.onUserLeaveHint()
   }
 
   override fun onWindowFocusChanged(hasFocus: Boolean) {
     super.onWindowFocusChanged(hasFocus)
-    callbacksRelay.accept(createWindowFocusEvent(hasFocus))
+
+    val result = callbacksFlow.tryEmit(createWindowFocusEvent(hasFocus))
+    val s = ""
```
psteiger: I don't get it. Left over from local testing?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Presenter.kt
```diff
@@ -31,23 +33,22 @@ import org.checkerframework.checker.guieffect.qual.UIEffect
  * it becomes unclear where you should write your bussiness logic.
  */
 abstract class Presenter : ScopeProvider {
-  private val behaviorRelay = BehaviorRelay.create<PresenterEvent>()
-  private val lifecycleRelay = behaviorRelay.toSerialized()
+  private val lifecycleFlow = MutableSharedFlow<PresenterEvent>()
```
davissuber, line 36: Does this one need the "replay = 1" passed in the constructor as well?

psteiger, line 36: Shouldn't this be `(1, 0, DROP_OLDEST)`? Without a buffer, `tryEmit` will always fail.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerScopeProvider.kt
```diff
@@ -19,12 +19,18 @@ import com.uber.autodispose.ScopeProvider
...
 /** [ScopeProvider] for [Worker] instances.  */
 open class WorkerScopeProvider internal constructor(
-  private val workerLifecycleObservable: Observable<WorkerEvent>
+  private val workerLifecycle: Flow<WorkerEvent>
 ) : ScopeProvider {
+  internal constructor(workerLifecycleObservable: Observable<WorkerEvent>) : this(workerLifecycleObservable.asFlow())
+
   override fun requestScope(): CompletableSource {
-    return workerLifecycleObservable.skip(1).firstElement().ignoreElement()
+    return workerLifecycle.drop(1).asObservable().firstElement().ignoreElement()
```
psteiger, line 34: Alternatively we could use
```kotlin
rxCompletable { workerLifecycle.take(2).collect() }
```

But not sure if there's a real advantage here, I just feel it's less conversions. (Flow -> Completable vs Flow -> Observable -> Maybe -> Completable)

psteiger (changes requested):

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
tyvsmith: junk, already removed.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Presenter.kt
tyvsmith: Yes, left over from testing on StateFlow

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
tyvsmith: Yeah, this is a reasonable suggestion. This might save some overhead on autodispose calling these apis.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerScopeProvider.kt
tyvsmith: As discussed offline, we need to specify DIspatcher.UNCONFINED here, as otherwise it's default and tests fail.

psteiger (changes requested):

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
psteiger: More experimentation: Notice that with current Rx implementation, the issue of calling `unbinder.unbind()` on a worker thread is the same: we'll end up calling `worker.onStop` on the caller worker thread:

```kotlin
    val unbinder = PublishSubject.create<Int>()
    Observable.merge(Observable.just(1), unbinder).subscribe {
        println("$it ${Thread.currentThread().name}")
    }
    Dispatchers.Default {
        delay(100)
        unbinder.onNext(2)
    }
```
Output:
```
1 main
2 DefaultDispatcher-worker-1
```

If we use `observeOn` before `subscribe`, this fixes it, but introduces unwanted asynchronicity, hence my proposal to forbid unbinding from a worker thread.

tyvsmith: Regarding 1, that's what I observed as well, and was thinking we'd need to use create a coroutinecontext on the spot. We can take one in as optional in the params for testing, but the current usage does "feel" leaky since it subscribes directly, but it looks like it ends up getting coupled to the parents lifecycle through the internal logic inside it.

Regarding 2, @FranAguilera has done some research on workers binding on main thread vs background and has a proposal to make an explicit API change. WIth his proposed changes, we'd need both options.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -40,31 +37,30 @@ import com.uber.rib.core.lifecycle.ActivityLifecycleEvent
...
+  private val lifecycleFlow = MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
+  private val callbacksFlow = MutableSharedFlow<ActivityCallbackEvent>(0, 1, BufferOverflow.DROP_OLDEST)
+  private val lifecycleObservable = lifecycleFlow.asObservable()
+  private val callbacksObservable = callbacksFlow.asObservable()
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun lifecycle(): Observable<ActivityLifecycleEvent> {
-    return lifecycleRelay.hide()
-  }
+  override fun lifecycle() = lifecycleObservable
```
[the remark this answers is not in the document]
tyvsmith: We do get this from the interface it inherits, I don't have a strong opinion on making this explicit besides being consistent.

@jbarr21 What are your thoughts on adhering to explicit-api=strict with RIBs? We get other API protections from Japicmp? Is that enough?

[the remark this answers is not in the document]
jbarr21: i usually err on the side of explicitness

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -41,30 +38,32 @@ import com.uber.rib.core.lifecycle.ActivityLifecycleEvent.Companion.create
...
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.rx2.asObservable
 
 /** Base implementation for all VIP [android.app.Activity]s. */
 abstract class RibActivity : CoreAppCompatActivity(), ActivityStarter, LifecycleScopeProvider<ActivityLifecycleEvent>, RxActivityEvents {
   private var router: ViewRouter<*, *>? = null
-  private val lifecycleBehaviorRelay = BehaviorRelay.create<ActivityLifecycleEvent>()
-  private val lifecycleRelay: Relay<ActivityLifecycleEvent> = lifecycleBehaviorRelay.toSerialized()
-  private val callbacksRelay = PublishRelay.create<ActivityCallbackEvent>().toSerialized()
+  private val lifecycleFlow = MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
```
jbarr21, line 48: do publish & behavior relays (which act slightly different) both get mapped to the same flow type?

android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/InteractorAndRouterTest.kt
```diff
@@ -102,6 +102,7 @@ class InteractorAndRouterTest {
     }
     val router = TestRouterA(parentInteractor, component)
     val parentObserver = RecordingObserver<InteractorEvent>()
+    parentInteractor.lifecycle()
```
psteiger, line 105: Left-over? Otherwise LGTM

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
tyvsmith: In this case yes.

https://github.com/Kotlin/kotlinx.coroutines/issues/2515 is an interesting discussion that sheds some light on this, but the TLDR is that StateFlow is more or less a very opinionated SharedFlow configuration and a simplified API to access the value. As our usecase doesn't meet that opinionated impl (required a default value), we need to go the route of configuring the SharedFlow for the behavior that we need of a BehaviorRelay without default value.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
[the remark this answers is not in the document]
tyvsmith: Made these explicit for now. We should take a pass and enable the compiler flag for explict-api in a separate PR.

jbarr21 (approved):

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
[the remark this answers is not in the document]
tyvsmith: https://github.com/uber/RIBs/issues/545

psteiger (approved):

PR #547, opened 2023-03-31
Introduce `RibCoroutineWorker`

### LINKED ISSUE #546: Introduce a coroutine-based `Worker`.
Current `Worker` was built with `Rx` in mind, so it takes in a `lifecycle: WorkerScopeProvider` and `onStart` is non-suspending.

If we want to launch coroutines on `onStart`, it's possible with the interop between lifecycle (`ScopeProvider`) and `CoroutineScope`:

```kotlin
fun onStart(lifecycle: WorkerScopeProvider) {
  lifecycle.asCoroutineScope().launch { 
    // coroutine body    
  }
}
```

However, this solution has some shortcomings, both in library maintenance and API usage:

1. We can't call suspend functions directly from `onStart`, instead we need to launch a coroutine.
2. (1) implies binder can't know whether the code we want to run on `onStart` has completed, as asynchronicity was introduced.
3. We have to maintain mappers between RIB components lifecycle (`Interactor`, `Presenter`) to worker lifecycle (`Observable<WorkerEvent>`), which itself is a new lifecycle component. But `CoroutineScope` already provides the `lifecycle` concept we need.
4. It's not easy to bind a `Worker` with a custom lifecycle. Current binder supports `Interactor` and `Presenter`. It's easy to create custom `CoroutineScope` with custom lifecycle, though, which could be used as the lifecycle for `Worker`. This unlocks worker to be used on more scenarios than the pre-defined lifecycles.

As the RIBs project moves toward a coroutine-first direction, I suggest we introduce a new kind of `Worker` (new to maintain retro-compatibility) that takes in the form (tentative):

A suspend `onStart`:
```
suspend fun onStart(scope: CoroutineScope)
```

A more flexible binder that requires only a `CoroutineScope`:
```
fun CoroutineScope.bind(worker: Worker)
```

The binder API proposed above easily supports any lifecycle-aware (`ScopeProvider`) RIB component we have. For example, in interactors we could just:
```
override fun didBecomeActive() {
  coroutineScope.bind(worker)
}
```

psteiger:
Introduces `RibCoroutineWorker`, a worker version with a suspending `onStart, a (non-suspending) `onStop` that takes in the termination cause, and more flexible binding capabilities.

## API comparison

The `RibCoroutineWorker ` contract is as follows:
```kotlin
public fun interface RibCoroutineWorker {
  /** Called when worker is started. Children coroutines can be launched in [scope]. */
  public suspend fun onStart(scope: CoroutineScope)

  /** Called when the worker is stopped with the given [cause]. Should be fast, non-blocking and should not throw. */
  public fun onStop(cause: Throwable) {}
}
```
As it can be seen, `RibCoroutineWorker` interface is very similar to `Worker`'s, but 
1. It has a suspend `onStart` that takes in a `CoroutineScope` (as opposed to non-suspend `onStart` that takes in a `WorkerScopeProvider`)
2. `onStop` receives the cancellation cause (parent scope cancelled, or worker manually cancelled by unbinder, or `onStart` method threw an exception, meaning binding failed).

Binding to a worker requires a `CoroutineScope` instead of an `Interactor` or `Presenter`. Any `ScopeProvider` can bind a worker through the `ScopeProvider.coroutineScope` extension.

```kotlin
public fun CoroutineScope.bind(worker: RibCoroutineWorker): BindWorkerHandle
```

```kotlin
override fun didBecomeActive() {
  coroutineScope.bind(worker)
}
```

Binding operation is non-suspending and is asynchronous, but unlike `Worker` binding API, it can be `joined` using the resulting `BindWorkerHandle`.

```kotlin
/**
 * A handle to interact with [RibCoroutineWorker] binding job. This handle implements [Job], which refers to the
 * completion of [RibCoroutineWorker.onStart]. It can be [joined][join] to make sure
 * [RibCoroutineWorker.onStart] finished.
 */
public sealed interface BindWorkerHandle : Job {
  /** Unbinds the worker. */
  public fun unbind(): Job
}
```

This allows for (hopefully) useful feedback to the caller component:
1. It can, on a coroutine, `join`ed to wait for `onStart` to finish. Exceptions are propagated to the parent.
2. `unbind()` is also asynchronous but returns an `Job` which can also be `joined`, so caller can know unbinding is completed:

Example:
```kotlin
override fun didBecomeActive() {
  // start worker binding work asynchronously
  val handle: BindWorkerHandle = coroutineScope.bind(worker)
  coroutineScope.launch {
    handle.join() // onStart finished. Failures are propagated upwards.
    handle.unbind().join() // onStop finished
  } 
}
```

## Alternative API considered

1. Suspending `onStop`

`onStop` is non-suspending. Suspending `onStop` was considered, but it's troublesome: by the time it is called, the scope may have completed already (for example, parent scope was cancelled externally). So `onStop` would fail at suspension points.

2. Suspending `onStart` that does not take in a `CoroutineScope`.
We could have a suspending `onStart` that does not take in a scope, but can open a scope itself to launch children coroutines:
```kotlin
suspend fun onStart() {
  coroutineScope {
    someFlow.launchIn(this)
  }
}
```

This API is fine, but if non-terminating flows (such as `SharedFlow` or `StateFlow`) are collected inside `onStart`, `onStart` would never terminate successfully, and binder caller would have no way to know if `onStart` finished itself but still has children coroutines running (flow collectors). It would probably be a source of surprise that `BindWorkerHandle.join()` would never return.

3. Non-suspending `onStart` that takes in a `CoroutineScope`
This would be the most similar to `Worker` API and, as the `Worker`, would require launching new coroutines every time we want to use a suspending function. Because all suspending work would be done in children coroutines, binder would have no way to know the suspending operations completed (which is the current form of `Worker`).

4. Suspending binding and unbinding operations.
This suspending API for binding and unbinding would minimize the Handles API (as joining, awaiting, etc would not be necessary), but would require to launch a coroutine just for binding in Interactors (repetitive and cumbersome to use), when we may not care about waiting for `onStart` to complete. 
```kotlin
override fun didBecomeActive() {
  coroutineScope.launch {
    coroutineScope.bind(worker) // suspending, returns only after onStart completes
  }
}
```
Also, it is confusing to have a function that both takes in a `CoroutineScope` (implying it launches work asynchronously) and also is suspending (implying that by the time it returns, all pending work has been done).

## Validation

Validated with a local build the following behaviors:
1. `onStart` is called by default on `RibDispatchers.Default` context.
2.  Overriding dispatcher in `bind` works.
3. `onStop` is always called on `onStart` failure, scope cancellation, and on manual unbinding.
4. `onStop` execution context is defined by the `CoroutineDispatcher`.
5. Best-effort synchronicity is respected in immediate dispatchers such as `Dispatchers.Main.immediate`. 
6. Verified `BindWorkerHandle` `join` and `unbind`.

Also introduced unit tests to verify most of the above behavior.

Fixes #546

### DIFF
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
@@ -15,6 +15,7 @@
  */
 package com.uber.rib.core
 
+import com.uber.autodispose.coroutinesinterop.asScopeProvider
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableJob
 import kotlinx.coroutines.CoroutineDispatcher
@@ -28,6 +29,7 @@ import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.withContext
 import kotlin.coroutines.ContinuationInterceptor
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
@@ -150,3 +152,45 @@ private fun CompletableJob.cancelOrCompleteExceptionally(throwable: Throwable) {
     else -> completeExceptionally(throwable)
   }
 }
+
+// ---- RibCoroutineWorker <-> Worker adapters ---- //
+
+/** Converts a [Worker] to a [RibCoroutineWorker]. */
+public fun Worker.asRibCoroutineWorker(): RibCoroutineWorker = WorkerToRibCoroutineWorkerAdapter(this)
+
+/** Converts a [RibCoroutineWorker] to a [Worker]. */
+@JvmOverloads
+public fun RibCoroutineWorker.asWorker(
+  dispatcher: CoroutineDispatcher? = null,
+): Worker = RibCoroutineWorkerToWorkerAdapter(this, dispatcher)
+
+internal open class WorkerToRibCoroutineWorkerAdapter(private val worker: Worker) : RibCoroutineWorker {
+  override suspend fun onStart(scope: CoroutineScope) {
+    withContext(worker.coroutineDispatcher) {
+      worker.onStart(scope.asWorkerScopeProvider())
+    }
+  }
+
+  override fun onStop(cause: Throwable): Unit = worker.onStop()
+}
+
+internal open class RibCoroutineWorkerToWorkerAdapter(
+  private val ribCoroutineWorker: RibCoroutineWorker,
+  private val dispatcher: CoroutineDispatcher?,
+) : Worker {
+  override val coroutineDispatcher: CoroutineDispatcher get() = dispatcher ?: super.coroutineDispatcher
+
+  override fun onStart(lifecycle: WorkerScopeProvider) {
+    // We can start it undispatched because Worker binder will already call `onStart` in correct context,
+    // but we still want to pass in `coroutineDispatcher` to resume from suspensions in `onStart` in correct context.
+    lifecycle.coroutineScope.launch(coroutineDispatcher, start = CoroutineStart.UNDISPATCHED) {
+      supervisorScope {
+        ribCoroutineWorker.onStart(this)
+      }
+    }
+  }
+
+  override fun onStop(): Unit = ribCoroutineWorker.onStop(CancellationException("Worker is unbinding."))
+}
+
+private fun CoroutineScope.asWorkerScopeProvider() = WorkerScopeProvider(asScopeProvider())
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
@@ -1,0 +1,152 @@
+/*
+ * Copyright (C) 2023. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.core
+
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CompletableJob
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.awaitCancellation
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.job
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.supervisorScope
+import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.coroutineContext
+import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
+import kotlin.coroutines.intrinsics.intercepted
+import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
+import kotlin.coroutines.resume
+
+/** A manager or helper class bound to a [CoroutineScope] by using a binder like [bind]. */
+public fun interface RibCoroutineWorker {
+  /** Called when worker is started. Children coroutines can be launched in [scope]. */
+  public suspend fun onStart(scope: CoroutineScope)
+
+  /** Called when the worker is stopped with the given [cause]. Should be fast, non-blocking and should not throw. */
+  @JvmDefault
+  public fun onStop(cause: Throwable) {}
+}
+
+// ---- Binder ---- //
+
+/**
+ * A handle to interact with [RibCoroutineWorker] binding job. This handle implements [Job], which refers to the
+ * completion of [RibCoroutineWorker.onStart]. It can be [joined][join] to make sure `onStart` finished. Note that
+ * children coroutines launched in the [CoroutineScope] passed on to `onStart` are not waited: worker is considered
+ * bound when `onStart` finishes.
+ */
+public sealed interface BindWorkerHandle : Job {
+  /** Unbinds the worker. */
+  public fun unbind(): Job
+}
+
+private class BindWorkerHandleImpl(
+  bindJob: Job,
+  private val unbindJob: Job,
+) : BindWorkerHandle, Job by bindJob {
+  override fun unbind(): Job {
+    unbindJob.cancel("Worker was manually unbound.")
+    return unbindJob
+  }
+}
+
+/**
+ * Binds [worker] in a scope that is a child of the [CoroutineScope] receiver.
+ *
+ * The binding operation runs [RibCoroutineWorker.onStart] in a context inherited from the [CoroutineScope] receiver, but
+ * with additional [context] elements that is, by default, [RibDispatchers.Default]. This makes the worker run on the
+ * default dispatcher by default. Pass in [EmptyCoroutineContext] instead if you want the worker to not override the
+ * dispatcher in the scope (if any), or pass in a custom dispatcher as [context] to specify a different dispatcher. If
+ * there is no dispatcher in [CoroutineScope] nor in [context], [RibDispatchers.Default] is used.
+ *
+ * The scope passed on to [RibCoroutineWorker.onStart] as a parameter is a child scope of the [CoroutineScope] receiver,
+ * but with the additional [context] elements and a [SupervisorJob].
+ *
+ * Binding a worker is an asynchronous operation. To ensure [RibCoroutineWorker.onStart] is finished, callers can
+ * [join][BindWorkerHandle.join] the resulting [BindWorkerHandle] when in a coroutine:
+ *
+ * ```
+ * val handle = coroutineScope.bind(worker)
+ * handle.join() // wait for onStart to finish
+ * ```
+ */
+@JvmOverloads
+public fun CoroutineScope.bind(
+  worker: RibCoroutineWorker,
+  context: CoroutineContext = RibDispatchers.Default,
+): BindWorkerHandle {
+  // A job that completes once worker's onStart completes
+  var bindJob: CompletableJob? = null
+  // Start the bind coroutine undispatched to make sure we set bindJob synchronously.
+  val unbindJob = launch(context, start = CoroutineStart.UNDISPATCHED) {
+    bindJob = Job(coroutineContext.job).also {
+      // Cancel `unbindJob` if `bindJob` has cancelled. This is important to abort `onStart` if `bindJob`
+      // gets cancelled externally.
+      // Note that in case of `bindJob` failure (e.g. `onStart` throws), `unbindJob` will already fail by means of
+      // structured concurrency, but that does not happen on normal cancellation.
+      // This `bindJob` cancellation upon the `unbindJob` cancellation is also already set through structured concurrency.
+      it.invokeOnCompletion { throwable -> if (throwable is CancellationException) cancel(throwable) }
+    }
+    // Verify cancellation and dispatch (if needed) using dispatcher in context instead of continuing on current thread.
+    dispatchIfNeeded()
+    bindAndAwaitCancellation(worker, bindJob!!)
+  }
+  // We know deferred is non-null because binding coroutine was started undispatched.
+  return BindWorkerHandleImpl(bindJob!!, unbindJob)
+}
+
+private suspend inline fun dispatchIfNeeded() {
+  suspendCoroutineUninterceptedOrReturn sc@{ cont ->
+    val context = cont.context
+    val dispatcher = context[ContinuationInterceptor] as CoroutineDispatcher
+    if (!dispatcher.isDispatchNeeded(context)) return@sc Unit
+    // Coroutine was not in the right context -- we'll dispatch.
+    context.ensureActive()
+    cont.intercepted().resume(Unit)
+    COROUTINE_SUSPENDED
+  }
+  coroutineContext.ensureActive() // Don't continue if coroutine was cancelled after returning from dispatch.
+}
+
+@Suppress("TooGenericExceptionCaught") // Exception is not swallowed
+private suspend fun bindAndAwaitCancellation(worker: RibCoroutineWorker, bindJob: CompletableJob) {
+  try {
+    supervisorScope {
+      worker.onStart(this)
+      ensureActive() // Before completing the binding deferred, make sure the binding coroutine is active.
+      bindJob.complete()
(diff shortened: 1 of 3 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -34,3 +44,39 @@ interface Worker {
   @JvmDefault
   fun onStop() {}
 }
+
+// ---- CoroutineWorker <-> Worker adapters ----//
+
+public fun Worker.asCoroutineWorker(): CoroutineWorker = WorkerToCoroutineWorkerAdapter(this)
```
FranAguilera, line 50: This seems a bit weird to me. Given that we adding a whole new "Rib" coroutine worker interface, this definition within a "rib" Worker shouldn't have to know anything about CoroutineWorker

android/libraries/rib-coroutines/src/main/kotlin/com/uber/rib/core/CoroutineWorker.kt
```diff
@@ -0,0 +1,148 @@
...
+import kotlinx.coroutines.supervisorScope
+import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.coroutineContext
+import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
+import kotlin.coroutines.intrinsics.intercepted
+import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
+import kotlin.coroutines.resume
+
+/** A manager or helper class bound to a [CoroutineScope] by using a binder like [bind]. */
+public interface CoroutineWorker {
```
FranAguilera, line 40: To name this differently to be clear this is RIB related worker as this could bring confusion with https://www.google.com/url?q=https://developer.android.com/reference/kotlin/androidx/work/CoroutineWorker&sa=D&source=docs&ust=1681257103494195&usg=AOvVaw0uf0BmqQexs8Wx5rHJC2vC

android/libraries/rib-coroutines/src/test/kotlin/com/uber/rib/core/CoroutineWorkerTest.kt
```diff
@@ -0,0 +1,232 @@
...
+  fun doOnStart(block: suspend () -> Unit) {
+    _doOnStart = block
+  }
+
+  override suspend fun onStart(scope: CoroutineScope) {
+    onStartStarted = true
+    onStartThread = Thread.currentThread()
+    try {
+      scope.launch { awaitCancellation() }
+      delay(ON_START_DELAY_DURATION_MILLIS)
```
FranAguilera, line 220: why this delay?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
psteiger: This can surely be in a different file. It can't be in the `rib-coroutine` module, though, as `rib-base` contains `Worker` and depends on `rib-coroutine` and not the other way around. `rib-coroutine` doesn't know about `Worker`.

psteiger: Just want to add that, even though it can surely be in a different file, not sure if it *should*. 

In fact, I don't think all of the following components should be defined in a file on their own, as they currently are: `Worker`, `WorkerBinder`, `WorkerScopeProvider`, `WorkerUnbinder`. This one-class per file (or one interface per file) is reminiscent of Java, but in Kotlin it's fine (even preferable) to put all closely related components (classes, interfaces, top-level functions, etc) on a single file.

android/libraries/rib-coroutines/src/test/kotlin/com/uber/rib/core/CoroutineWorkerTest.kt
psteiger: I think I initially wanted to test something when the binding was in the state of started, but not yet finished (onStart suspended). Going over current tests, I don't think it is of any importance, so I'll just remove it. Thanks!

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
psteiger: Moved `RibCoroutineWorker` to `ribs-base` and adapters to `RibCoroutineWorker.kt`

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -0,0 +1,197 @@
...
+      onStart(this)
+      ensureActive() // Before completing the binding deferred, make sure the binding coroutine is active.
+      deferred.complete(Unit)
+      awaitCancellation() // Never returns normally, so we are sure an exception will be caught.
+    }
+  } catch (t: Throwable) {
+    t
+  }
+  deferred.cancelOrCompleteExceptionally(terminationCause) // No effect if deferred is already completed.
+  onStop(terminationCause)
```
SergeySmykovskyi, line 140: Looks like we will never receive runtime crashes by this logic. I think will be really hard to catch the bug in large project, if your changes broking someone's else Worker. I would recommend to throw all Exceptions caused by `onStart` by default and introduce some `Cause` sealed class for cleaner handling : 
```kotlin
sealed class Cause {
  object Lifecycle : Cause()
  class Exception(val t: Throwable) : Cause()
}
```

psteiger: That's an interesting criticism, @SergeySmykovskyi. However, given that binding is an asynchronous operation in this design (`bind` is non-suspend, called outside of a coroutine), I don't see how we can synchronously throw! Exposing exceptions to the user in coroutines is done by `await`-ing in coroutines. 

To address this criticism, `bind` would need to be `suspend` itself (so it could throw into the caller coroutine), but it gives rise to another set of issues as discussed in the design -- mainly that it's poor practice to both launch asynchronous operations (non-terminating coroutines like shared flow collection) and suspend in a `suspend fun`. When a function is `suspend`, callers expect that by the time it returns it has already done all of its work. Alternatively, if we did not pass in a `CoroutineScope` on `onStart` (so users would need to open its own scope to collect non-terminating flow, using for example `supervisorScope { }`), the binding operation would never terminate. 

Do you have a concrete suggestion on how to tackle this?

SergeySmykovskyi: Might need deeper investigation but I saw it as simple solution. Default implementation of `onStop` :
```kotlin
  @JvmDefault
  @CallSuper // Required to be sure that even overridden `onStop`  will throw unhandled exception (something like in `RXJava`)
  public fun onStop(cause: Cause) {
    if (cause is Cause.Exception) {
      handleException(cause.t)
    }
  }

  @JvmDefault
  public fun handleException(t: Throwable) {
    throw t 
  }
```

psteiger: @SergeySmykovskyi I've taken your concern into consideration and decided to change the binding job from `async` (exposes exception to user by means of `.await()`) into `launch` (automatically propagates exception to parent).

This will cause the parent scope to fail with exception, and if parent is a supervisor without a `CoroutineExceptionHandler`, it will crash the application *after* delivering the exception to `onStop`.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -0,0 +1,188 @@
...
+  var bindJob: CompletableJob? = null
+  // Start the bind coroutine undispatched to make sure we set bindJob synchronously.
+  val unbindJob = launch(context, start = CoroutineStart.UNDISPATCHED) {
+    bindJob = Job(coroutineContext.job).also {
+      // Cancel unbindJob if `bindJob` has completed exceptionally. This is important to abort `onStart` if `bindJob`
+      // gets cancelled externally.
+      // Notice that in most cases, `unbindJob` will already fail when `bindJob` fails (`onStart` throws) by means of
+      // structured concurrency, but that does not happen on normal cancellation.
+      // This `bindJob` cancellation upon the `unbindJob` cancellation is also already set through structured concurrency.
+      it.invokeOnCompletion { t -> if (t != null) cancel("Bind job has completed exceptionally", t) }
```
FranAguilera, line 104: nit: rename `t` for `throwable`

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
psteiger: Validated latest changes on internal codebase. Throwing on `onStart` crashes the app. Unbinding (normal cancellation) doesn't.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -0,0 +1,187 @@
...
+ * Binding a worker is an asynchronous operation. To ensure [RibCoroutineWorker.onStart] is finished, callers can
+ * [join][BindWorkerHandle.join] the resulting [BindWorkerHandle] when in a coroutine:
+ *
+ * ```
+ * val handle = coroutineScope.bind(worker)
+ * handle.join() // wait for onStart to finish
+ * ```
+ */
+@JvmOverloads
+public fun CoroutineScope.bind(
```
FranAguilera, line 87: Btw, what do you think if we add 2 new methods to ease existing binding of [Workers] off main thread (when called on didBecome interactor)

Internally often times interactors will bind a big list a list of workers via plugin factory. This should ease the binding for those instances and easily switch them from UI to RibDispatchers.Default without major changes

Was thinking something like this

public fun CoroutineScope.bindRibWorkers(workers:List<Worker>, context: CoroutineContext = RibDispatchers.Default){
  for(worker in workers){
    this.bindRibWorker(worker, context)
  }
}

public fun CoroutineScope.bindRibWorker(
  worker: Worker,
  context: CoroutineContext = RibDispatchers.Default): BindWorkerHandle{

  val ribCoroutineWorker = worker.asRibCoroutineWorker()
  return bind(ribCoroutineWorker, context)
}

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -0,0 +1,187 @@
...
+  override fun unbind(): Job {
+    unbindJob.cancel("Worker was manually unbound.")
+    return unbindJob
+  }
+}
+
+// ---- RibCoroutineWorker <-> Worker adapters ---- //
+
+/** Converts a [Worker] to a [RibCoroutineWorker]. */
+public fun Worker.asRibCoroutineWorker(): RibCoroutineWorker = WorkerToRibCoroutineWorkerAdapter(this)
```
FranAguilera, line 165: sorry missed this earlier. Wondering if we should be adding this now given that old presidio Worker now defines coroutineDispatcher too (the resulting threading type may be confusing to the user).Also at this moment we would allow binding existing worker via RibCoroutine binder without a monitoring option

psteiger: > Wondering if we should be adding this now given that old presidio Worker now defines coroutineDispatcher too

I think yes, because we will still have (internal) components that expect a `Worker` in the signature so in the short term we can expect some `Worker` <-> `RibCoroutineWorker` conversions to happen. It's an easy way to plug-in the new contract into existing components expecting the old contract without requiring refactoring.

> Also at this moment we would allow binding existing worker via RibCoroutine binder without a monitoring option

Yes, but it's a general problem with coroutines we need to tackle. It's not as trivial as adapting the current implementation for `Worker` since monitoring `suspend fun` is not as simple as `measureTimeMillis` the function call because of the suspension capability (what if a function suspends for a long time? There's no harm in that as it's not occupying thread resources).

I'm planning on kicking off a general coroutine monitoring proposal soon, to be pluggable into our dispatchers so we can monitor not only workers, but basically any coroutine running on the system.

FranAguilera: Got it, thanks a lot for the clarification. It make senses to me.

The only thing that could be a bit unclear and didn't add to the previous WorkerBinder PR was the capability to specify the coroutineDispatcher at WorkerBinder.bind methods given it could bring some confusion to the resulting dispatcher at Worker. e.g. WorkerBinder.bind with `RibDispatchers.IO` and `Worker` coroutineDispatcher override with `RibDispatchers.Main`.This would require some tweaks on the binding logic to always honor the dispatcher defined on the Worker but maybe will be good to bring this back so Worker still has the last preference on defining its dispatcher given binding a list of workers on WorkerBinder.bind call (likely to support pluginPoint)  is a frequent pattern on our monorepo

In the adapter case `Worker.asRibCoroutineWorker(): RibCoroutineWorker`, there could be a `Worker` being built to have an expectation to run on UI thread (overriding default Unconfined explicitly with Main). If the resulting `RibCoroutineWorker` is bound in different dispatcher, it won't honor the dispatcher defined at original Worker

FranAguilera: Actually, the custom dispatcher in Worker is probably not a bit deal for Worker<>RibCoroutineWorker conversion. 

If there is an intention to convert a Worker to RibCouroutineDispatcher is kind of implicit that now you have better control of the dispatchers being used at binding

FranAguilera (approved):

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
psteiger: Got it! I figure we can probably have a corresponding property in the `RibCoroutineWorker` to specify the dispatcher, remove the `context` parameter in `CoroutineScope.bind`, and ignore the dispatcher in `CoroutineScope` (if any). That would make both impls very similar (and easily convertible between themselves without losing semantics) and also I think it's a simpler mental model when there's a single place to decide the dispatcher.

Thanks for the feedback! Will re-ask for review once patched.

SergeySmykovskyi (approved):

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
psteiger: Hey, revisiting this.

On further thought:

1. It does make sense to provide a `val coroutineDispatcher` in `Worker` but it does not make sense in `RibCoroutineWorker`, because `RibCoroutineWorker` can always override the context with `withContext` (since `onStart` is `suspend`). There's no need for it to be a `val` of the interface. This also solves the issue of binding multiple workers at once: every worker can become the SoT of execution context with `withContext`.
```kotlin
override suspend fun onStart(scope: CoroutineScope) = withContext(RibDispatchers.Main) {
  // ...
}
```

Synchronous `Worker` cannot use `withContext` on `onStart`, so it needs to be a `val` on there.

2. It stills makes sense for specifying a default dispatcher in `CoroutineScope.bind`, because if the worker does not care on which dispatcher it runs, *we want it to run on RibDispatchers.Default by default*.

3. There's a fair point on `Worker.asRibCoroutineWorker()` conversion: `Worker` provides a `val coroutineDispatcher` to specify the context of execution of `onStart`. *We must still respect that after conversion*. So I'm patching the conversor with:
```kotlin
  override suspend fun onStart(scope: CoroutineScope) {
    withContext(worker.coroutineDispatcher) {
      worker.onStart(scope.asWorkerScopeProvider())
    }
  }
```

FranAguilera: Great point and approach for honoring Worker coroutine dispatcher on the conversions. Looks great to me

PR #550, opened 2023-04-14
Redesign RouterAndState to avoid router caching

SergeySmykovskyi:
**Description**:
This change resolves possible memory leaks produced by storing `router` instances in the `RouterAndState` objects. 

**Related issue(s)**: NONE reported yet

**Leak canary:**
Screenshot 2023-04-14 at 10 55 41 AM

### DIFF
--- a/android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
+++ b/android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
@@ -16,6 +16,8 @@
 package com.uber.rib.core
 
 import androidx.annotation.VisibleForTesting
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 
 /**
  * Internal class for keeping track of a navigation stack.
@@ -31,45 +33,37 @@ internal class RouterAndState<R : Router<*>, StateT : RouterNavigatorState>(
   detachTransition: RouterNavigator.DetachTransition<R, StateT>?,
   @get:VisibleForTesting val forceRouterCaching: Boolean = false
 ) {
-  @get:Synchronized
-  @set:Synchronized
-  private var _router: R? = null
+  private val routerAccessor = SafeRouterAccessor(attachTransition::buildRouter)
 
   /**
    * Gets or creates the [Router] associated with this state. Router will be destroyed after
    * [RouterNavigator.DetachCallback.onPostDetachFromHost] and if [StateT] is cacheable.
-   *
-   * @return [Router]
    */
-  internal val router: R
-    @Synchronized
-    get() =
-      _router
-        ?: attachTransition.buildRouter().apply {
-          log("Router ${this@apply.javaClass.simpleName} was created")
-          _router = this@apply
-        }
+  internal val router: R by routerAccessor::router
 
   internal fun willAttachToHost(previousState: StateT?, isPush: Boolean) =
-    attachTransition.willAttachToHost(router, previousState, state, isPush)
+    routerAccessor.safeOperation { safeRouter ->
+      attachTransition.willAttachToHost(safeRouter, previousState, state, isPush)
+    }
 
-  internal fun willDetachFromHost(
-    newState: StateT?,
-    isPush: Boolean
-  ) = detachCallback.willDetachFromHost(router, state, newState, isPush)
+  internal fun willDetachFromHost(newState: StateT?, isPush: Boolean) =
+    routerAccessor.safeOperation { safeRouter ->
+      detachCallback.willDetachFromHost(safeRouter, state, newState, isPush)
+    }
 
   internal fun onPostDetachFromHost(newState: StateT?, isPush: Boolean) =
-    detachCallback.onPostDetachFromHost(router, newState, isPush)
+    routerAccessor.safeOperation { safeRouter ->
+      detachCallback.onPostDetachFromHost(safeRouter, newState, isPush)
+    }
 
-  /**
-   * Gets the [RouterNavigator.DetachCallback] associated with this state.
-   *
-   * @return [RouterNavigator.DetachCallback]
-   */
   private val detachCallback: RouterNavigator.DetachCallback<R, StateT> by lazy {
     RouterDestroyerCallbackWrapper(
       baseCallback = wrapDetachTransitionIfNeed(detachTransition),
-      onDestroy = ::destroyRouterIfNeed
+      onDestroy = {
+        routerAccessor.safeConditionalDestroy(state) {
+          !forceRouterCaching && !state.isCacheable()
+        }
+      }
     )
   }
 
@@ -80,23 +74,9 @@ internal class RouterAndState<R : Router<*>, StateT : RouterNavigatorState>(
       ?: detachTransition?.let { DetachCallbackWrapper(it) }
   }
 
-  private fun destroyRouterIfNeed() {
-    if (!forceRouterCaching && !state.isCacheable()) {
-      val routerName = _router?.javaClass?.simpleName
-      _router = null
-      if (routerName != null) {
-        log("Destroying router $routerName was destroyed")
-      } else {
-        log("Router of ${state.stateName()} state already destroyed")
-      }
-    }
-  }
-
   /**
-   * Wrapper class to wrap [RouterNavigator.DetachTransition] calls into the new
-   * [RouterNavigator.DetachCallback] format.
-   *
-   * @param transitionCallback Base transaction
+   * Wrapper class to wrap [transitionCallback] calls into the new [RouterNavigator.DetachCallback]
+   * format.
    */
   private inner class DetachCallbackWrapper(
     private val transitionCallback: RouterNavigator.DetachTransition<R, StateT>
@@ -107,11 +87,13 @@ internal class RouterAndState<R : Router<*>, StateT : RouterNavigatorState>(
       previousState: StateT,
       newState: StateT?,
       isPush: Boolean
-    ) {
-      transitionCallback.willDetachFromHost(router, previousState, newState, isPush)
-    }
+    ) = transitionCallback.willDetachFromHost(router, previousState, newState, isPush)
   }
 
+  /**
+   * Wrapper class to wrap [RouterNavigator.DetachCallback] and call [onDestroy] after
+   * [RouterNavigator.DetachCallback.onPostDetachFromHost]
+   */
   private inner class RouterDestroyerCallbackWrapper(
     private val baseCallback: RouterNavigator.DetachCallback<R, StateT>?,
     private val onDestroy: () -> Unit
@@ -129,6 +111,37 @@ internal class RouterAndState<R : Router<*>, StateT : RouterNavigatorState>(
     override fun onPostDetachFromHost(router: R, newState: StateT?, isPush: Boolean) {
       baseCallback?.onPostDetachFromHost(router, newState, isPush)
       onDestroy.invoke()
+    }
+  }
+
+  private class SafeRouterAccessor<R : Router<*>>(
+    private val routerBuilder: () -> R
+  ) {
+    private val lock = ReentrantLock()
+    private var _router: R? = null
+
+    val router: R
+      get() = lock.withLock {
+        _router ?: routerBuilder().let { newRouter ->
+          log("Router ${newRouter.javaClass.simpleName} was created")
+          _router = newRouter
+          newRouter
+        }
+      }
+
+    fun safeOperation(operation: (R) -> Unit) = lock.withLock {
+      operation(router)
+    }
+
+    fun safeConditionalDestroy(state: RouterNavigatorState, condition: () -> Boolean) = lock.withLock {
+      if (condition.invoke()) {
+        _router?.javaClass?.simpleName?.let { routerName ->
+          log("Destroying router $routerName was destroyed")
+          _router = null
+        } ?: run {
+          log("Router of ${state.stateName()} state already destroyed")
+        }
+      }
     }
   }
 
--- a/android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
+++ b/android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
@@ -1,0 +1,141 @@
+/*
+ * Copyright (C) 2023. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.core
+
+import androidx.annotation.VisibleForTesting
+
+/**
+ * Internal class for keeping track of a navigation stack.
+ *
+ * @param state The associated state
+ * @param attachTransition The [RouterNavigator.AttachTransition] associated with this state.
+ * @param detachTransition The [RouterNavigator.DetachTransition] associated with this state.
+ * @param forceRouterCaching Override [RouterNavigatorState.isCacheable] behavior
+ */
+internal class RouterAndState<R : Router<*>, StateT : RouterNavigatorState>(
+  val state: StateT,
+  private val attachTransition: RouterNavigator.AttachTransition<R, StateT>,
+  detachTransition: RouterNavigator.DetachTransition<R, StateT>?,
+  @get:VisibleForTesting val forceRouterCaching: Boolean = false
+) {
+  @get:Synchronized
+  @set:Synchronized
+  private var _router: R? = null
+
+  /**
+   * Gets or creates the [Router] associated with this state. Router will be destroyed after
+   * [RouterNavigator.DetachCallback.onPostDetachFromHost] and if [StateT] is cacheable.
+   *
+   * @return [Router]
+   */
+  internal val router: R
+    @Synchronized
+    get() =
+      _router
+        ?: attachTransition.buildRouter().apply {
+          log("Router ${this@apply.javaClass.simpleName} was created")
+          _router = this@apply
+        }
+
+  internal fun willAttachToHost(previousState: StateT?, isPush: Boolean) =
+    attachTransition.willAttachToHost(router, previousState, state, isPush)
+
+  internal fun willDetachFromHost(
+    newState: StateT?,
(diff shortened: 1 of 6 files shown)


android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
```diff
@@ -275,7 +260,8 @@ interface RouterNavigator<StateT : RouterNavigatorState> {
     }
 
     /**
-     * Notifies the consumer that the [StackRouterNavigator] has detached the supplied [ ]. Consumers can complete any post detachment behavior here.
+     * Notifies the consumer that the [StackRouterNavigator] has detached the supplied [ ].
+     * Consumers can complete any post detachment behavior here.
      *
      * @param router [Router]
```
psteiger, line 266: IMO we can remove all those @param that add nothing to the reader. They're ubiquitous in JavaDoc but not much common in KDoc.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
```diff
@@ -95,18 +94,16 @@ interface RouterNavigator<StateT : RouterNavigatorState> {
...
    *
    * NOTE: This will retain the Riblet in memory until it is popped or detached by a push with
    * certain flags.
    *
    * @param newState to switch to.
    * @param attachTransition method to attach child router.
    * @param detachTransition method to clean up child router when removed.
-   * @param <R> router type to detach.
-   </R> */
+   * @param <R> router type to detach. </R>
```
psteiger, line 105: I'd remove those `@param` that add nothing to the reader. I'd KDoc it like the following:
```kotlin
/**
 * Detaches the current state using [detachTransition] and attaches [newState] using [attachTransition]. 
 * The resulting stack depends on the [flag].
 */
```

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
```diff
@@ -118,18 +115,16 @@ interface RouterNavigator<StateT : RouterNavigatorState> {
    * Switch to a new state - this will switch out children if the state is not the current active
    * state already.
    *
-   *
-   * NOTE: This will retain the Riblet in memory until it is popped. To push transient, riblets,
-   * use [RouterNavigator.pushTransientState]
-   *
+   * NOTE: This will retain the Riblet in memory until it is popped. To push transient, riblets, use
```
psteiger, line 118: This is probably need some adjusting after your changes. We'll not retain Riblet in memory anymore. Should we deprecate the DEFAULT flag?

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/StackRouterNavigator.kt
```diff
@@ -16,24 +16,33 @@
...
+ * @param hostRouter Host router
+ * @param forceRouterCaching Override [RouterNavigatorState.isCacheable] behavior
  */
-open class StackRouterNavigator<StateT : RouterNavigatorState>(private val hostRouter: Router<*>) : RouterNavigator<StateT> {
+open class StackRouterNavigator<StateT : RouterNavigatorState>
+@JvmOverloads
+constructor(
+  private val hostRouter: Router<*>,
+  // Enabled by default for safe migration to newer lib version
+  private val forceRouterCaching: Boolean = true
```
psteiger, line 35: I believe adding a default parameter in constructor makes it a binary-incompatible change. We should use two secondary constructors instead.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/StackRouterNavigator.kt
```diff
@@ -133,27 +142,32 @@ open class StackRouterNavigator<StateT : RouterNavigatorState>(private val hostR
...
+          navigationStack.push(newRouterAndState)
+        }
+
       RouterNavigator.Flag.NEW_TASK_REPLACE -> {
         detachAll()
         newRouterAndState = buildNewState(newState, attachTransition, detachTransition)
         attachInternal(currentRouterAndState, newRouterAndState, true)
         navigationStack.push(newRouterAndState)
       }
+
```
psteiger, line 170: Not a fan of those extra spaces

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/StackRouterNavigator.kt
```diff
@@ -198,13 +211,18 @@ open class StackRouterNavigator<StateT : RouterNavigatorState>(private val hostR
     navigationStack.clear()
   }
 
-  private fun buildNewState(
+  @Suppress("UNCHECKED_CAST")
+  private fun <R : Router<*>> buildNewState(
```
psteiger, line 215: Why UNCHECKED_CAST? Try to use covariance instead

android/libraries/rib-router-navigator/src/test/kotlin/com/uber/rib/core/RouterAndStateTest.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    val router1 = routerAndState.router
+    routerAndState.detachCallback.onPostDetachFromHost(router1, mock(), false)
+    val router2 = routerAndState.router
+
+    Truth.assertThat(router1).isSameAs(router2)
+  }
+
+  @Test
+  fun router___StateIsCacheable_ForcedToCache___reuseRouterOnNextAttach() {
```
psteiger, line 88: I don't think we use this `__` convention on this lib

android/libraries/rib-router-navigator/src/test/kotlin/com/uber/rib/core/StackRouterNavigatorTest.kt
```diff
@@ -455,10 +438,7 @@ class StackRouterNavigatorTest {
     routerNavigator.pushState(TestState.STATE_1, attachTransition1, detachTransition1)
     routerNavigator.pushState(TestState.STATE_2, attachTransition2, detachTransition2)
     routerNavigator.pushState(
-      TestState.STATE_3,
-      RouterNavigator.Flag.NEW_TASK,
-      attachTransition3,
-      detachTransition3
+      TestState.STATE_3, RouterNavigator.Flag.NEW_TASK, attachTransition3, detachTransition3
```
psteiger, line 441: IMO, changes that just format code should be avoided. This makes the PR harder to review.

psteiger (changes requested):

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/StackRouterNavigator.kt
psteiger: Actually it is in accordance of coding conventions

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/StackRouterNavigator.kt
SergeySmykovskyi: This constructor has `@JvmOverloads` annotation which will generate single-parameter constructor for usages in java code. If I understand you comment correctly.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
```diff
@@ -0,0 +1,130 @@
...
+  val attachTransition: RouterNavigator.AttachTransition<Router<*>, StateT>,
+  detachTransition: RouterNavigator.DetachTransition<Router<*>, StateT>?,
+  @get:VisibleForTesting val forceRouterCaching: Boolean = false
+) {
+  @get:Synchronized
+  @set:Synchronized
+  private var innerRouter: Router<*>? = null
+
+  /**
+   * Gets or creates the [Router] associated with this state. Router will be destroyed after
```
jbarr21, line 39: will the router (thus the whole RIB) lose any state since this is no longer a strong reference?

jbarr21 (approved):

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
psteiger: I think we must double-check not only going forward but **going back** on the stack for features that use this stacking with flag DEFAULT, to ensure no regressions are introduced.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/StackRouterNavigator.kt
SergeySmykovskyi: Tried to do so before, but with covariance we loose router type needed in  the `attachInternal` and `detachInternal` methods. I working on code change to avoid this casts, will update PR soon.

android/libraries/rib-router-navigator/src/test/kotlin/com/uber/rib/core/RouterAndStateTest.kt
SergeySmykovskyi: Make sense, working on change.

android/libraries/rib-router-navigator/src/test/kotlin/com/uber/rib/core/StackRouterNavigatorTest.kt
SergeySmykovskyi: Will remove similar changes

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
SergeySmykovskyi: I will modify comment for the cleaner understanding. "Memory" mentioned in the comment means navigation stack memory () not the physical heap memory. Changes in this PR doesn't affect any navigation logic, so `DEFAULT` flag  still will do the same things.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
SergeySmykovskyi: Correct, the `Router` instance will lose object state (changes in private properties). The right way to save any state is to use `Bundle` provided in `Router.dispatchAttach` (for state recovery) and `Router.saveInstanceState` (for the state saving). For the retro compatibility I've added new [isCacheable](https://github.com/uber/RIBs/pull/550/files#diff-5fb0dee6f8224afb283316fb92377bf3291abc13b217b0b8507f8733d7d08561) property, which is `false` by default (as recommended) + extended the `StackRouterNavigator` constructor by `forceRouterCaching` parameter which will override `isCacheable` by `true`, so after updating library we will have old behavior by default.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
SergeySmykovskyi: I will remove this and similar changes from this PR and will create separate one to do that.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
```diff
@@ -0,0 +1,141 @@
...
+ * @param detachTransition The [RouterNavigator.DetachTransition] associated with this state.
+ * @param forceRouterCaching Override [RouterNavigatorState.isCacheable] behavior
+ */
+internal class RouterAndState<R : Router<*>, StateT : RouterNavigatorState>(
+  val state: StateT,
+  private val attachTransition: RouterNavigator.AttachTransition<R, StateT>,
+  detachTransition: RouterNavigator.DetachTransition<R, StateT>?,
+  @get:VisibleForTesting val forceRouterCaching: Boolean = false
+) {
+  @get:Synchronized
```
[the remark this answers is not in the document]
SergeySmykovskyi: Not sure what did you mean by "whole object". I've marked only the `get` and `set` of `_router` by `Synchronized`, this annotation will generate (if not exist) private `$lock` object for locking, if we will have any additional method which doesn't touching `_router`, it will not be synchronized.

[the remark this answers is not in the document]
SergeySmykovskyi: Also, I can add small internal `synchronizedProperty` delegate, so it will look like : 
`var _router : R? by synchronizedProperty()`

tyvsmith:
looks good overall, what's the plan to start the roll-out of this? It looks like we'll need to handle the bundle per usage, so requires a call-site migration for each one?

SergeySmykovskyi:
> @tyvsmith  We'll need to handle the bundle per usage, so requires a call-site migration for each one?
We have `RouterNavigatorFactory` which can be used for the migration

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterNavigator.kt
SergeySmykovskyi: I will do it in separate PR, please check newest version of PR, it doesn't include changes like this one

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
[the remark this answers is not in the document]
psteiger: Just did some verifications
```kotlin
public class Test {
  @get:Synchronized
  public val a: Int = 3
}
```
Translates to
```java
public final class Test {
   private final int a = 3;

   public final synchronized int getA() {
      return this.a;
   }
}
```
So `@Synchronized` translates to a Java [synchronized method](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html`, which synchronizes on the instance of `RouterAndState`.

What I meant with my comment is that if we have another field and mark that other field as `@Synchronized`, the getter for this field would also be synchronized on the class instance, and hence two threads, one accessing `property A` and other accessing `property B`, would have to wait one for the other when not needed. But that's not the case on your impl.

Nevertheless, I see some issues in the code. For example:
```kotlin
  private fun destroyRouterIfNeed() {
    if (!forceRouterCaching && !state.isCacheable()) {
      val routerName = _router?.javaClass?.simpleName
      _router = null
      if (routerName != null) {
        log("Destroying router $routerName was destroyed")
      } else {
        log("Router of ${state.stateName()} state already destroyed")
      }
    }
  }
  ```

This method reads `_router`, then writes to `_router`. But the lock in between those operations is lost, and anything could happen to this field in the meantime. In other words, the whole operation is not atomic, though reading and writing to the property individually are atomic. 

I'd suggest either synchronizing all methods that touch the property for having fully atomic operations, or probably better yet, wrap the property in  [AtomicReference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native.concurrent/-atomic-reference/). With the latter, above method could be written as:

```kotlin
private fun destroyRouterIfNeed() {
  if (!forceRouterCaching && !state.isCacheable()) {
    // notice the block can run multiple times, so should be side effect free
    val previous = _router.getAndUpdate { null }
    val routerName = current?.javaClass?.simpleName
    if (routerName != null) {
      log("Destroying router $routerName was destroyed")
    } else {
      log("Router of ${state.stateName()} state already destroyed")
    }
  }
}
```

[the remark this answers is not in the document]
SergeySmykovskyi: Good point ! Already redesigned for thread safe usage + covered by unit test (test fails without new logic thread-safe)

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }
+
+  private class SafeRouterAccessor<R : Router<*>>(
+    private val routerBuilder: () -> R
+  ) {
+    private val lock = ReentrantLock()
+    private var _router: R? = null
+
+    fun router(): R = lock.withLock {
```
psteiger, line 124: Would probably use `val router: R get() =` instead

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
```diff
@@ -0,0 +1,153 @@
...
+      _router ?: routerBuilder.invoke().apply {
```
psteiger, line 125: It probably makes sense to use `.also { newRouter ->` instead of apply here, since we're not applying any new stuff to the newly create router (instead we're attaching it to `_router`)

android/libraries/rib-router-navigator/src/test/kotlin/com/uber/rib/core/RouterAndStateTest.kt
```diff
@@ -0,0 +1,193 @@
...
+          routerAndState.onPostDetachFromHost(null, false)
+        }
+        finishedThreadsCount.addAndGet(1)
+      }
+    }
+
+    while (finishedThreadsCount.get() < threadsCount) {
+      // Waiting until all threads finish their work
+      Thread.sleep(10)
+    }
```
psteiger, line 188: A countdown latch is a better solution than randomly sleeping 10ms in a tight loop:
```kotlin
    val threadsCount = 10
    val latch = CountDownLatch(threadsCount)
    // Create 10 threads with 1000 routers usage flows inside each
    repeat(threadsCount) {
      thread {
        repeat(1000) {
          routerAndState.willAttachToHost(null, false)
          routerAndState.willDetachFromHost(null, false)
          routerAndState.onPostDetachFromHost(null, false)
        }
        latch.countDown()
      }
    }
    latch.await()
```

psteiger (approved):
Looks good! Approved with minor (nit) comments.

android/libraries/rib-router-navigator/src/main/kotlin/com/uber/rib/core/RouterAndState.kt
SergeySmykovskyi: Make sense, I've used  `apply` because it's not requires to return any value from it but 'm agree the `let` is better for code readability

android/libraries/rib-router-navigator/src/test/kotlin/com/uber/rib/core/RouterAndStateTest.kt
SergeySmykovskyi: Nice, didn't know that `CountDownLatch ` exist.

tyvsmith (approved):

Issue #545, opened 2023-03-27
Enable explict-api=strict on Kotlin libraries.


tyvsmith:

psteiger:
I believe that now with #544, #548, #552, #561 merged, all Kotlin modules have been set to use strict explicit API mode, so I'm closing this issue.

PR #552, opened 2023-04-14
Reduce `Rx` <-> `Coroutines` interop and allow unconfined coroutines to run eagerly inside Worker's onStart

### LINKED ISSUE #551: Allow unconfined coroutines to run without entering an event loop inside Worker's onStart.
The recent Rx <-> Coroutines migration (f15ec67) introduced a backwards-incompatible change.

Assume the following:

```kotlin
private class FakeInteractor(private val parentJob: CompletableJob) : Interactor<Unit, Router<FakeInteractor>>() {
  val worker = object : Worker {
    override fun onStart(lifecycle: WorkerScopeProvider) {
      val subscription = lifecycle().subscribe {
        // 0
      }
      // 1
    }
  }

  override fun didBecomeActive(savedInstanceState: Bundle?) {
    bind(this, worker)
  }
}
```

Before the migration, we could expect that `lifecycle().subscribe` would be entered eagerly (`0` runs before `1`).

After the migration, it is *not* entered eagerly (`1` runs before `0`)

The root cause for this is:

1. Interactor lifecycle is now backed by a flow, and is converted to `Observable` by `lifecycleFlow.asObservable()`
2. `asObservable` is cold, but when subscribed to it launches a coroutine using `Dispatchers.Unconfined` to collect the flow and emit to the `Observable` it creates.
3. Worker binder subscribes to the interactor lifecycle observable (defined as `lifecycleFlow.asObservable()`), so it launches a coroutine under `Dispatchers.Unconfined` to bind the worker (call `onStart`).
4. `Worker.onStart` might also launch coroutines in unconfined dispatcher. Even subscribing to `Observable` converted from `Flow` (such as the lifecycle passed on by parameter to `onStart`) triggers this. 
5. Consequently, **nested coroutines using Dispatchers.Unconfined** have formed. [This results in the usage of an event loop](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-unconfined.html).

psteiger:
Reducing RxJava <-> Coroutines interop minimizes the chances for surprising behavior. A flow converted to Rx converted to a Flow converted to Rx (...) will run multiple nested coroutines using `Dispatchers.Undispatched` (as per `Flow.asObservable()`), which will form event loops that we want to avoid.

This PR also changes worker binding implementation (`bind(Interactor, Worker)`) to use the non-converted Interactor lifecycle, and to launch binding coroutine in unconfined dispatcher while still starting it undispatched. 

### Tests
This PR introduces a test that passes without f15ec672eab55c5b7a72bab60d625dfab564d33c, fails with it, and passes again after these changes.

Also, validated the core flow in internal codebase that previously crashed with latest `main`. 

Fixes #551

### DIFF
--- a/android/libraries/rib-base/build.gradle
+++ b/android/libraries/rib-base/build.gradle
@@ -43,7 +43,7 @@ dependencies {
     implementation deps.apt.javaxInject
 
     implementation deps.uber.autodisposeCoroutines
-    implementation deps.kotlin.coroutinesAndroid
+    implementation deps.kotlin.coroutines
     implementation deps.kotlin.coroutinesRx2
 
     implementation project(":libraries:rib-coroutines")
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
@@ -20,12 +20,12 @@ import androidx.annotation.VisibleForTesting
 import com.uber.autodispose.lifecycle.CorrespondingEventsFunction
 import com.uber.autodispose.lifecycle.LifecycleEndedException
 import com.uber.autodispose.lifecycle.LifecycleScopeProvider
-import com.uber.autodispose.lifecycle.LifecycleScopes
 import com.uber.rib.core.lifecycle.InteractorEvent
 import io.reactivex.CompletableSource
 import io.reactivex.Observable
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.rx2.asObservable
 import javax.inject.Inject
 import kotlin.properties.ReadWriteProperty
@@ -41,8 +41,11 @@ public abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvide
   @Inject
   public lateinit var injectedPresenter: P
   internal var actualPresenter: P? = null
-  private val lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  private val lifecycleObservable = lifecycleFlow.asObservable()
+  private val _lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
+  // We expose lifecycleFlow internally to avoid some unnecessary Rx-Coroutine interop. Use it instead of `lifecycle()`.
+  @get:JvmSynthetic
+  internal val lifecycleFlow: SharedFlow<InteractorEvent> get() = _lifecycleFlow
+  private val lifecycleObservable = _lifecycleFlow.asObservable()
 
   private val routerDelegate = InitOnceProperty<R>()
   /** @return the router for this interactor. */
@@ -59,7 +62,7 @@ public abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvide
   override fun lifecycle(): Observable<InteractorEvent> = lifecycleObservable
 
   /** @return true if the controller is attached, false if not. */
-  override fun isAttached(): Boolean = lifecycleFlow.replayCache.lastOrNull() == InteractorEvent.ACTIVE
+  override fun isAttached(): Boolean = _lifecycleFlow.replayCache.lastOrNull() == InteractorEvent.ACTIVE
 
   /**
    * Called when attached. The presenter will automatically be added when this happens.
@@ -95,7 +98,7 @@ public abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvide
   protected open fun onSaveInstanceState(outState: Bundle) {}
 
   public open fun dispatchAttach(savedInstanceState: Bundle?) {
-    lifecycleFlow.tryEmit(InteractorEvent.ACTIVE)
+    _lifecycleFlow.tryEmit(InteractorEvent.ACTIVE)
     (getPresenter() as? Presenter)?.dispatchLoad()
     didBecomeActive(savedInstanceState)
   }
@@ -103,7 +106,7 @@ public abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvide
   public open fun dispatchDetach(): P {
     (getPresenter() as? Presenter)?.dispatchUnload()
     willResignActive()
-    lifecycleFlow.tryEmit(InteractorEvent.INACTIVE)
+    _lifecycleFlow.tryEmit(InteractorEvent.INACTIVE)
     return getPresenter()
   }
 
@@ -137,13 +140,9 @@ public abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvide
     return LIFECYCLE_MAP_FUNCTION
   }
 
-  override fun peekLifecycle(): InteractorEvent? {
-    return lifecycleFlow.replayCache.lastOrNull()
-  }
+  override fun peekLifecycle(): InteractorEvent? = _lifecycleFlow.replayCache.lastOrNull()
 
-  final override fun requestScope(): CompletableSource {
-    return LifecycleScopes.resolveScopeFromLifecycle(this)
-  }
+  final override fun requestScope(): CompletableSource = _lifecycleFlow.asScopeCompletable(lifecycleRange)
 
   private inner class InitOnceProperty<T> : ReadWriteProperty<Any, T> {
     private var backingField: T? = null
@@ -165,6 +164,9 @@ public abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvide
   }
 
   public companion object {
+    @get:JvmSynthetic
+    internal val lifecycleRange = InteractorEvent.ACTIVE..InteractorEvent.INACTIVE
+
     private val LIFECYCLE_MAP_FUNCTION = CorrespondingEventsFunction { interactorEvent: InteractorEvent ->
       when (interactorEvent) {
         InteractorEvent.ACTIVE -> return@CorrespondingEventsFunction InteractorEvent.INACTIVE
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -21,12 +21,13 @@ import com.uber.rib.core.lifecycle.InteractorEvent
 import com.uber.rib.core.lifecycle.PresenterEvent
 import com.uber.rib.core.lifecycle.WorkerEvent
 import io.reactivex.Observable
-import kotlinx.coroutines.channels.BufferOverflow
-import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.merge
-import kotlinx.coroutines.flow.transformWhile
-import kotlinx.coroutines.rx2.asFlow
-import kotlinx.coroutines.rx2.asObservable
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.takeWhile
+import kotlinx.coroutines.launch
 
 /** Helper class to bind to an interactor's lifecycle to translate it to a [Worker] lifecycle. */
 public object WorkerBinder {
@@ -40,9 +41,8 @@ public object WorkerBinder {
    * @return [WorkerUnbinder] to unbind [Worker&#39;s][Worker] lifecycle.
    */
   @JvmStatic
-  public fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder {
-    return bind(mapInteractorLifecycleToWorker(interactor.lifecycle()), worker)
-  }
+  public fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder =
+    worker.bind(interactor.lifecycleFlow, Interactor.lifecycleRange)
 
   /**
    * Bind a list of workers (ie. a manager or any other class that needs an interactor's lifecycle)
@@ -68,9 +68,8 @@ public object WorkerBinder {
    * @return [WorkerUnbinder] to unbind [Worker&#39;s][Worker] lifecycle.
    */
   @JvmStatic
-  public fun bind(presenter: Presenter, worker: Worker): WorkerUnbinder {
-    return bind(mapPresenterLifecycleToWorker(presenter.lifecycle()), worker)
-  }
+  public fun bind(presenter: Presenter, worker: Worker): WorkerUnbinder =
+    worker.bind(presenter.lifecycleFlow, Presenter.lifecycleRange)
 
   /**
    * Bind a list of workers (ie. a manager or any other class that needs an presenter's lifecycle)
@@ -90,14 +89,11 @@ public object WorkerBinder {
   @JvmStatic
   @VisibleForTesting
   public fun bind(mappedLifecycle: Observable<WorkerEvent>, worker: Worker): WorkerUnbinder {
-    val unbindFlow = MutableSharedFlow<WorkerEvent>(0, 1, BufferOverflow.DROP_OLDEST)
-
-    val workerLifecycle = merge(mappedLifecycle.asFlow(), unbindFlow)
-      .transformWhile { emit(it) ; it != WorkerEvent.STOP }
-      .asObservable()
-
-    bindToWorkerLifecycle(workerLifecycle, worker)
-    return WorkerUnbinder { unbindFlow.tryEmit(WorkerEvent.STOP) }
+    val disposable = mappedLifecycle
+      .takeWhile { it != WorkerEvent.STOP }
+      .doFinally { worker.onStop() }
+      .subscribe { worker.onStart(WorkerScopeProvider(mappedLifecycle)) }
+    return WorkerUnbinder(disposable::dispose)
   }
 
   @JvmStatic
@@ -150,6 +146,13 @@ public object WorkerBinder {
    * @param worker the class that wants to be informed when to start and stop doing work
    */
   @JvmStatic
+  @Deprecated(
+    message = """
+      This method is unsafe because it assumes that the 'workerLifecycle' Observable completes after it
+      emits 'WorkerEvent.STOP'. If it does not complete, the subscription will leak.
+    """,
+    replaceWith = ReplaceWith("bind(workerLifecycle, worker)")
+  )
   public fun bindToWorkerLifecycle(
     workerLifecycle: Observable<WorkerEvent>,
     worker: Worker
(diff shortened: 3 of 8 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -39,9 +42,20 @@ object WorkerBinder {
...
    */
+  @OptIn(DelicateCoroutinesApi::class)
   @JvmStatic
-  open fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder {
-    return bind(mapInteractorLifecycleToWorker(interactor.lifecycle()), worker)
+  public fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder {
+    val workerEvents = interactor.lifecycleFlow.map { it.asWorkerEvent() }
+    // We need to launch in Unconfined dispatcher, AND start undispatched.
+    // If we only use Unconfined, and `onStart` launches some coroutine on Unconfined, coroutine will be put on an event loop.
+    val job = GlobalScope.launch(RibDispatchers.Unconfined, start = CoroutineStart.UNDISPATCHED) {
```
SergeySmykovskyi, line 51: This looks a little dangerous for me.  From my knowledge we expecting that `onStart` and `onStop` will be called on main/UI thread. I think it's OK to collect events on `Dispatchers.Unconfined` but we should switch to main in `onEach` and `onCompletion` blocks.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -39,9 +42,20 @@ object WorkerBinder {
...
+      workerEvents
+        .takeWhile { it == WorkerEvent.START }
+        .onEach { worker.onStart(WorkerScopeProvider(workerEvents)) }
```
SergeySmykovskyi, line 54: Let add `distinctUntilChanged` between just to make sure that onStart will not be called multiple times somewhy.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -89,15 +103,12 @@ object WorkerBinder {
...
-
-    val workerLifecycle = merge(mappedLifecycle.asFlow(), unbindFlow)
-      .transformWhile { emit(it) ; it != WorkerEvent.STOP }
-      .asObservable()
-
-    bindToWorkerLifecycle(workerLifecycle, worker)
-    return WorkerUnbinder { unbindFlow.tryEmit(WorkerEvent.STOP) }
+  public fun bind(mappedLifecycle: Observable<WorkerEvent>, worker: Worker): WorkerUnbinder {
+    val d = mappedLifecycle
+      .takeWhile { it == WorkerEvent.START }
```
SergeySmykovskyi, line 108: Same `distinctUntilChanged`

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerScopeProvider.kt
```diff
@@ -20,20 +20,30 @@ import com.uber.rib.core.lifecycle.WorkerEvent
...
+
+  internal constructor(workerLifecycleObservable: Observable<WorkerEvent>) {
+    this.workerLifecycleFlow = null
+    this.workerLifecycleObservable = workerLifecycleObservable
+  }
+
+  override fun requestScope(): CompletableSource = when {
+    workerLifecycleFlow != null -> workerLifecycleFlow.asScopeCompletable(WorkerEvent.STOP)
+    workerLifecycleObservable != null -> workerLifecycleObservable.skip(1).firstElement().ignoreElement()
+    else -> error("WorkerScopeProvider must have either a lifecycle Flow or lifecycle Observable.")
```
SergeySmykovskyi, line 47: Why we need `public` modifier ? And what about this style instead 
```suggestion
 /*
   * WorkerScopeProvider may be instantiated with a Flow or Observable of WorkerEvent.
   * We don't convert from one to another for `requestScope`. Instead, we have alternative implementations for each.
   * This is done to keep Rx <-> Coroutines interop to a minimum.
   */
sealed class WorkerScopeProvider : ScopeProvider {

internal class FlowBased(private val workerLifecycleFlow: Flow<WorkerEvent>) : WorkerScopeProvider() {
  override fun requestScope(): CompletableSource = 
    workerLifecycleFlow.asScopeCompletable(WorkerEvent.STOP)
  }

internal class ObservableBased(private val workerLifecycleObservable: Observable<WorkerEvent>) : WorkerScopeProvider() {
  override fun requestScope(): CompletableSource = 
    workerLifecycleObservable.skip(1).firstElement().ignoreElement()
  }
}
```

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interop.kt
```diff
@@ -0,0 +1,42 @@
...
+import io.reactivex.CompletableSource
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.rx2.rxCompletable
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+internal fun <T> Flow<T>.asScopeCompletable(
+  lastEvent: T,
```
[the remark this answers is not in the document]
psteiger: 1. I'm not ready to introduce it as public API.
2. In the long term, when 100% coroutine-based, it seems RIBs should not depend on AutoDispose at all (doesn't make sense if we already have `CoroutineScope` for lifecycle management). AutoDispose is useful for Rx.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
psteiger: In current `Worker` API, and still with this change, `onStart` and `onStop` will run on Main whenever `bind` is called on main. This commit does not change that. Note that calling `bind` on a worker thread already binds it (calls `onStart`) on a worker thread. There are 2 recent related PRs for defining the running thread for `onStart`. One introduces `RibCoroutineWorker` and other keeps `Worker` API but changes binding method.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
psteiger: I don't think it can happen. Actually, this should never happen, so maybe we better raise an exception if it ever happens, instead of just filtering it out, as it means programmer error. Thanks for this suggestion, I'll reflect on that a bit more. But notice current implementation does not filter out values whatsoever.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
psteiger: See response before

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerScopeProvider.kt
psteiger: > why we need public
We are going to make all libs strictly use explicit API, there's an issue opened for that (#545) and a PR (#548)

> what about this style instead
Good call, I actually thought of using a sealed class/interface when coding this. I'll change it.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
SergeySmykovskyi: The `Context.Unconfined` doc it's doesn't guarantee that `onStart` and `onStop` will be called on the same thread as `bind` method was called. I also cannot understand how `CoroutineStart.UNDISPATCHED` will resolve looping. Referencing to the doc again: 

> However, when the coroutine is resumed from suspension it is dispatched according to the CoroutineDispatcher in its context

The resume methods in our case (`onEach`, `onCompletion`) will use `Context.Unconfined` and that start parameter will have affect only on `takeWhile` predicate, so it should be the same behavior with this `start =` parameter or without it . Maybe I misunderstood something :)

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
SergeySmykovskyi: 👍

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
psteiger: Let's unwrap the flow (rough estimate):
```kotlin
flow {
  try { // onCompletion
    flow { // takeWhile 
      lifecycleFlow.collect {
        if (it == START) emit(it)
      }
    }.collect {
      emit(it)
    }
  } finally {
    worker.onStop()
  }
}.collect {
  worker.onStart()
}
```
Each inner emit immediately calls the lambda in the outer collect. Notice there is no suspension point whatsoever upon binding (except potentially `lifecycleFlow.collect` which I will discuss later), because `worker.onStart` and `worker.onStop` do not suspend. They will all be called synchronously in the context of the outermost collect.

Unconfined dispatcher docs state that it starts coroutine in the same call frame (similar to starting undispatched) and lets it resume from suspension on whatever thread called `continuation.resumeWith` after suspension. But we will never have any suspension when binding here, so this is irrelevant.

**How can it suspend?**

There's only one possibility in this code that makes it suspend upon binding: lifecycle flow has not emitted yet (lifecycle not started). It can be happen when trying to bind *before* `didBecomeActive`. In this corner case, we want `lifecycleFlow.collect` to resume immediately upon emission, and the **only** way to resume immediately, synchronously upon emission here is to do so on the thread that emitted the lifecycle event, which will be the main thread (see `Interactor`). For this resume immediately behavior, we need `Dispatches.Unconfined`.

**onStop**

I mentioned the flow will never suspend when binding (unless lifecycleFlow has no value), but what about unbinding? Obviously the coroutine will always suspend *after* binding, until the lifecycle terminates or user unbinds the worker.

How will resume happen in this case? If we use a dispatcher different than unconfined, the coroutine will not react immediately to the emission of the final lifecycle event. Instead the event delivery would be scheduled/dispatched. We rather have lifecycle collectors to react **immediately** to events.

**Why start both undispatched and in Unconfined**

I hope unconfined was justified by this point, but why also start undispatched?

android/libraries/rib-base/build.gradle
```diff
@@ -43,7 +43,7 @@ dependencies {
     implementation deps.apt.javaxInject
 
     implementation deps.uber.autodisposeCoroutines
-    implementation deps.kotlin.coroutinesAndroid
+    api deps.kotlin.coroutines
```
psteiger, line 46: This needs API now because we are exposing (`internal`) a `lifecycleFlow` which is a `SharedFlow`, so extenders must know `SharedFlow` at compile time (even though they can't access it).

Also, changed from `coroutinesAndroid` to `coroutines` -- this is not even an Android package, why are we using `coroutinesAndroid`?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerScopeProvider.kt
psteiger: Update: I did not make it sealed because it would break binary compatibility (e.g. mocks on `WorkerScopeProvider` would not work anymore), but nevertheless I made `WorkerScopeProvider` an interface and separated the implementation into two classes.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -41,8 +41,11 @@ abstract class Interactor<P : Any, R : Router<*>> : LifecycleScopeProvider<Inter
   @Inject
   lateinit var injectedPresenter: P
   internal var actualPresenter: P? = null
-  private val lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  private val lifecycleObservable = lifecycleFlow.asObservable()
+  private val _lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
```
FranAguilera, line 44: qq: why using underscore for instance class variable? The IDE will easily indicate this

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -39,9 +40,22 @@ object WorkerBinder {
...
-  open fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder {
-    return bind(mapInteractorLifecycleToWorker(interactor.lifecycle()), worker)
+  public fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder {
+    val workerEvents = interactor.lifecycleFlow.map { it.asWorkerEvent() }
+    /*
+     * We need `Dispatchers.Unconfined` to react immediately to lifecycle flow emissions, and we need
+     * `CoroutineStart.Undispatched` to prevent coroutines launched in `onStart` with `Dispatchers.Unconfined`
+     * from forming an event loop instead of starting eagerly.
+     */
+    val job = GlobalScope.launch(RibDispatchers.Unconfined, start = CoroutineStart.UNDISPATCHED) {
```
FranAguilera, line 52: related PR for updated WorkerBinder (still pending test and updating the worker duration listener) https://github.com/uber/RIBs/pull/553/

FranAguilera, line 52: why we need GlobalScope? wouldn't this introduce a risk to potential leaks?

psteiger: In theory it would, but we are using "takeWhile" on flow that makes it terminate once the interactor lifecycle terminates (so the coroutine also terminates)

The reason for not using `interactor.coroutineScope` is to keep retro-compatibily with calling bind before `didBecomeActive` which should be rare, but was supported by previous implementation nevertheless

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
psteiger: Just [Kotlin coding convention for private backing properties](https://kotlinlang.org/docs/coding-conventions.html#names-for-backing-properties)

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
SergeySmykovskyi: Great explanation, thanks for your time to write so detailed response 👍

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerScopeProvider.kt
psteiger: Update #2: Needed to change things a bit to maintain source compatibility. We need to still support `new WorkerScopeProvider(observable)` constructor. Final form is this:
```kotlin
public class WorkerScopeProvider private constructor(delegate: ScopeProvider): ScopeProvider by delegate {
  constructor(lifecycle: Flow<WorkerEvent>) : this(FlowScopeProvider(lifecycle))
  constructor(lifecycle: Observable<WorkerEvent>) : this(ObservableScopeProvider(lifecycle))
}

private class FlowScopeProvider(private val lifecycle: Flow<WorkerEvent>) : ScopeProvider {
  override fun requestScope(): CompletableSource = lifecycle.asScopeCompletable(WorkerEvent.STOP)
}

private class ObservableScopeProvider(private val lifecycle: Observable<WorkerEvent>) : ScopeProvider {
  override fun requestScope(): CompletableSource = lifecycle.skip(1).firstElement().ignoreElement()
}
```

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
FranAguilera: My bad, missed this was a backing properties. Thanks for sharing

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: I see, thanks for clarifying

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -89,15 +88,12 @@ object WorkerBinder {
...
-    val unbindFlow = MutableSharedFlow<WorkerEvent>(0, 1, BufferOverflow.DROP_OLDEST)
-
-    val workerLifecycle = merge(mappedLifecycle.asFlow(), unbindFlow)
-      .transformWhile { emit(it) ; it != WorkerEvent.STOP }
-      .asObservable()
-
-    bindToWorkerLifecycle(workerLifecycle, worker)
-    return WorkerUnbinder { unbindFlow.tryEmit(WorkerEvent.STOP) }
+  public fun bind(mappedLifecycle: Observable<WorkerEvent>, worker: Worker): WorkerUnbinder {
+    val d = mappedLifecycle
```
FranAguilera, line 92: optional nit: Maybe we could update `d` with a more meaningful name

psteiger: Good suggestion. A few chars more is worth the readability improvement.

android/libraries/rib-base/build.gradle
psteiger: Making the property `@JvmSynthetic` solves this issue -- reverted to `implementation`, but kept the change from `coroutinesAndroid` to `coroutines`

tyvsmith:
Please update description with test/verification steps taken.

psteiger:
> Please update description with test/verification steps taken.

Done!

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -90,14 +89,11 @@ public object WorkerBinder {
...
   public fun bind(mappedLifecycle: Observable<WorkerEvent>, worker: Worker): WorkerUnbinder {
-    val unbindFlow = MutableSharedFlow<WorkerEvent>(0, 1, BufferOverflow.DROP_OLDEST)
-
-    val workerLifecycle = merge(mappedLifecycle.asFlow(), unbindFlow)
-      .transformWhile { emit(it) ; it != WorkerEvent.STOP }
-      .asObservable()
-
-    bindToWorkerLifecycle(workerLifecycle, worker)
-    return WorkerUnbinder { unbindFlow.tryEmit(WorkerEvent.STOP) }
+    val disposable = mappedLifecycle
```
FranAguilera, line 92: I noticed there is also on line 153. Should this be updated accordingly as well? The params list seems very similar. It just differs on returning type 

@JvmStatic
  public fun bindToWorkerLifecycle(
    workerLifecycle: Observable<WorkerEvent>,
    worker: Worker
  ) {
    workerLifecycle.subscribe { workerEvent: WorkerEvent ->
      when (workerEvent) {
        WorkerEvent.START -> worker.onStart(WorkerScopeProvider(workerLifecycle.hide()))
        else -> worker.onStop()
      }
    }
  }

FranAguilera: Great points. In this case maybe this could be a good moment to add the deprecation notice there as well

psteiger: Deprecated it with:
```kotlin
  @Deprecated(
    message = """
      This method is unsafe because it assumes that the 'workerLifecycle' Observable completes after it
      emits 'WorkerEvent.STOP'. If it does not complete, subscription will be leaked.
    """,
    replaceWith = ReplaceWith("bind(workerLifecycle, worker)")
  )
```

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -68,9 +68,8 @@ public object WorkerBinder {
    * @return [WorkerUnbinder] to unbind [Worker&#39;s][Worker] lifecycle.
    */
   @JvmStatic
-  public fun bind(presenter: Presenter, worker: Worker): WorkerUnbinder {
-    return bind(mapPresenterLifecycleToWorker(presenter.lifecycle()), worker)
-  }
+  public fun bind(presenter: Presenter, worker: Worker): WorkerUnbinder =
```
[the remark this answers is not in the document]
FranAguilera: My bad, ignore this. I missed those internally call this bind method

FranAguilera (approved):
lgtm! thanks for fixing

PR #553, opened 2023-04-15
[Rib Worker] Specify CoroutineDispatcher for onStart/onStop and provide WorkerBinder monitoring option

FranAguilera:
# Issue

https://github.com/uber/RIBs/issues/555

# Depends on https://github.com/uber/RIBs/pull/552

# Goal

To be more explicit on the threading model where a RIB Worker.onStart/onStop will be running upon (will also include a monitoring option for Worker duration and the threading type they operate on)

**IMPORTANT: This PR only focus on updating existing production Rib Worker e.g. **com.uber.rib.core.Worker**. There is a new proposal on https://github.com/uber/RIBs/pull/547 in order to define a new RibCoroutineWorker that will run by default on background thread and will have onStart defined as a suspend function**

## Updating presidio Worker to allow specifying threading model

Worker will specify a new method in order to clarify on which thread a Worker onStart/onStop will be running upon 

Default implementation of this new method will be of  WORKER_BINDER_CALLER thread type in order to keep existing WorkerBind.bind as it (often times WorkerBinder.bind is called on UI thread)

```
@JvmDefault
public val coroutineDispatcher: CoroutineDispatcher
  get() = RibDispatchers.Unconfined
```

```
class FileCreationWorker: Worker {

 override val coroutineDispatcher: CoroutineDispatcher = RibDispatchers.IO

  override fun onStart(lifecycle: WorkerScopeProvider) {
     // Expensive IO operation here
  }

  override fun onStop() {
    // Also running on IO
  }
}
```

## Updating existing WorkerBinder.bind() internals

Update WorkerBinder internal methods to ensure onStart/onStop will happen on the specified Worker.threadingType()

Also a `RibWorkerBinderDurationListener` can be specified at `WorkerBinder` in order to get total binding duration of Worker.onStart/Worker.onStop

## Monitoring

Will provide monitoring mechanism to report total duration of Worker.onStart/onStop and which Threading type are they operating upon.

Sample reporting data for a sample expensive Worker

image

Reporting data will be something similar to this

```
data class BinderDuration(
  /**
   * Worker class name
   */
  val workerName: String,

  /**
   * Worker event type (START/STOP)
   */
  val workerEvent: WorkerEvent,

  /**
   * Threading Type where Worker will operate
   */
  val threadingType: ThreadingType,

  /**
   * Thread name where Worker.onStart/onStop was called
   */
  val threadName: String,

  /**
   * Total binding duration in milli of Worker.onStart/onStop
   */
  val totalBindingDurationMilli: Long
)
```

## Test plan + Verification

Tested WorkerBinder on Interactor's/Presenter by validating on demo libs. (Pending to add demo videos)

https://user-images.githubusercontent.com/4230063/233267153-81bf6c9f-176b-4604-8986-e2df1fbf3c64.mov

### DIFF
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
@@ -15,6 +15,8 @@
  */
 package com.uber.rib.core
 
+import kotlinx.coroutines.CoroutineDispatcher
+
 /**
  * Interface used when creating a manager or helper class that should be bound to an interactor's
  * lifecycle using a binder like [WorkerBinder]. The worker event is decoupled from the
@@ -22,6 +24,16 @@ package com.uber.rib.core
  * other lifecycles we're interested in.
  */
 public interface Worker {
+
+  /**
+   * Specifies in which [CoroutineDispatcher] WorkerBind.bind will be operating on
+   *
+   * NOTE: Default implementation will be using RibDispatchers.Unconfined to keep "caller" backward compatibility on existing usages
+   */
+  @JvmDefault
+  public val coroutineDispatcher: CoroutineDispatcher
+    get() = RibDispatchers.Unconfined
+
   /**
    * Called when worker is started.
    *
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -21,6 +21,7 @@ import com.uber.rib.core.lifecycle.InteractorEvent
 import com.uber.rib.core.lifecycle.PresenterEvent
 import com.uber.rib.core.lifecycle.WorkerEvent
 import io.reactivex.Observable
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.GlobalScope
@@ -28,9 +29,27 @@ import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.takeWhile
 import kotlinx.coroutines.launch
+import java.lang.ref.WeakReference
+import kotlin.system.measureTimeMillis
 
 /** Helper class to bind to an interactor's lifecycle to translate it to a [Worker] lifecycle. */
 public object WorkerBinder {
+
+  private var workerBinderListenerWeakRef: WeakReference<WorkerBinderListener>? = null
+
+  /**
+   * Initializes reporting of [WorkerBinderInfo] via [WorkerBinderListener]
+   *
+   * IMPORTANT: This should be called only once at early app scope to get proper monitoring of early worker
+   * being bound
+   *
+   */
+  @JvmStatic
+  public fun initializeMonitoring(workerBinderListener: WorkerBinderListener) {
+    this.workerBinderListenerWeakRef =
+      WeakReference<WorkerBinderListener>(workerBinderListener)
+  }
+
   /**
    * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
    * interactor's lifecycle events. Inject this class into your interactor and call this method on
@@ -42,7 +61,7 @@ public object WorkerBinder {
    */
   @JvmStatic
   public fun bind(interactor: Interactor<*, *>, worker: Worker): WorkerUnbinder =
-    worker.bind(interactor.lifecycleFlow, Interactor.lifecycleRange)
+    worker.bind(interactor.lifecycleFlow, Interactor.lifecycleRange, workerBinderListenerWeakRef)
 
   /**
    * Bind a list of workers (ie. a manager or any other class that needs an interactor's lifecycle)
@@ -69,7 +88,7 @@ public object WorkerBinder {
    */
   @JvmStatic
   public fun bind(presenter: Presenter, worker: Worker): WorkerUnbinder =
-    worker.bind(presenter.lifecycleFlow, Presenter.lifecycleRange)
+    worker.bind(presenter.lifecycleFlow, Presenter.lifecycleRange, workerBinderListenerWeakRef)
 
   /**
    * Bind a list of workers (ie. a manager or any other class that needs an presenter's lifecycle)
@@ -88,6 +107,14 @@ public object WorkerBinder {
 
   @JvmStatic
   @VisibleForTesting
+  @Deprecated(
+    message = """
+      This method doesn't support the [WorkerBinderThreadingType] defined at Worker. Due to this, the binding
+      will happen on the caller thread without a possibility to change the threading (even specifying a different WorkerThreadingType other than CALLER_THREAD)
+      It also doesn't provide information for WorkerBinderDuration when a [WorkerDurationMonitoringListener] is added
+    """,
+    replaceWith = ReplaceWith("bind(interactor, worker) or bind(presenter, worker)")
+  )
   public fun bind(mappedLifecycle: Observable<WorkerEvent>, worker: Worker): WorkerUnbinder {
     val disposable = mappedLifecycle
       .takeWhile { it != WorkerEvent.STOP }
@@ -166,10 +193,63 @@ public object WorkerBinder {
   }
 }
 
+/**
+ * Holds all relevant information for completed Worker.onStart/onStop actions.
+ * (e.g. Name of the Worker bound, duration of total onStart/onStop, thread name where onStart/onStop happens,etc)
+ */
+public data class WorkerBinderInfo(
+  /**
+   * Worker class name
+   */
+  val workerName: String,
+
+  /**
+   * Worker event type (START/STOP)
+   */
+  val workerEvent: WorkerEvent,
+
+  /**
+   * [CoroutineDispatcher] where [WorkerBinder.bind] will be operating upon
+   */
+  val coroutineDispatcher: CoroutineDispatcher,
+
+  /**
+   * Thread name where Worker.onStart/onStop was called.
+   *
+   * e.g. When [CoroutineDispatcher] is set as [RibDispatchers.Default] a sample threadName value would be similar to `DefaultDispatcher-worker-2`
+   */
+  val threadName: String,
+
+  /**
+   * Total binding duration in milliseconds of Worker.onStart/onStop
+   */
+  val totalBindingDurationMilli: Long
+)
+
+/**
+ * Reports total binding duration of Worker.onStart/onStop
+ */
+public fun interface WorkerBinderListener {
+
+  /**
+   * Reports all related Worker information via [WorkerBinderInfo] when onStart/onStop methods are completed
+   */
+  public fun onBindCompleted(
+    workerBinderInfo: WorkerBinderInfo
+  )
+}
+
 private fun <T : Comparable<T>> Worker.bind(
   lifecycle: SharedFlow<T>,
   lifecycleRange: ClosedRange<T>,
+  workerDurationListenerWeakRef: WeakReference<WorkerBinderListener>?
 ): WorkerUnbinder {
+
+  val coroutineStart = if (coroutineDispatcher == RibDispatchers.Unconfined) {
+    CoroutineStart.UNDISPATCHED
+  } else {
+    CoroutineStart.DEFAULT
+  }
   /*
    * We need `Dispatchers.Unconfined` to react immediately to lifecycle flow emissions, and we need
    * `CoroutineStart.Undispatched` to prevent coroutines launched in `onStart` with `Dispatchers.Unconfined`
@@ -178,11 +258,50 @@ private fun <T : Comparable<T>> Worker.bind(
    * GlobalScope won't leak the job, because the flow completes when lifecycle completes.
    */
   @OptIn(DelicateCoroutinesApi::class)
-  val job = GlobalScope.launch(RibDispatchers.Unconfined, start = CoroutineStart.UNDISPATCHED) {
+  val job = GlobalScope.launch(
+    coroutineDispatcher,
+    start = coroutineStart
+  ) {
     lifecycle
       .takeWhile { it < lifecycleRange.endInclusive }
-      .onCompletion { onStop() }
-      .collect { onStart(WorkerScopeProvider(lifecycle.asScopeProvider(lifecycleRange))) }
+      .onCompletion {
+        bindAndReportWorkerInfo(workerDurationListenerWeakRef, WorkerEvent.STOP) { onStop() }
+      }
+      .collect {
+        bindAndReportWorkerInfo(workerDurationListenerWeakRef, WorkerEvent.START) {
+          val workerScopeProvider = WorkerScopeProvider(lifecycle.asScopeProvider(lifecycleRange))
+          onStart(workerScopeProvider)
+        }
+      }
   }
   return WorkerUnbinder(job::cancel)
+}
+
+private inline fun Worker.bindAndReportWorkerInfo(
+  workerBinderListeners: WeakReference<WorkerBinderListener>?,
+  event: WorkerEvent,
+  workerBinderAction: Worker.() -> Unit
+) {
+  val duration = measureTimeMillis { workerBinderAction() }
+  workerBinderListeners?.reportWorkerBinderInfo(this, event, duration)
+}
+
+private fun WeakReference<WorkerBinderListener>.reportWorkerBinderInfo(
+  worker: Worker,
+  workerEvent: WorkerEvent,
+  totalBindingEventMilli: Long
+) {
+
+  val workerClassName = worker.javaClass.name
+  val currentThreadName = Thread.currentThread().name
+
+  val workerBinderInfo = WorkerBinderInfo(
+    workerClassName,
+    workerEvent,
+    worker.coroutineDispatcher,
+    currentThreadName,
+    totalBindingEventMilli
+  )
+
+  this@reportWorkerBinderInfo.get()?.onBindCompleted(workerBinderInfo)
 }
--- a/android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/WorkerBinderTest.kt
+++ b/android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/WorkerBinderTest.kt
@@ -24,14 +24,33 @@ import com.uber.rib.core.WorkerBinder.mapPresenterLifecycleToWorker
 import com.uber.rib.core.lifecycle.InteractorEvent
 import com.uber.rib.core.lifecycle.PresenterEvent
 import com.uber.rib.core.lifecycle.WorkerEvent
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.Before
+import org.junit.Rule
 import org.junit.Test
 import org.mockito.kotlin.any
+import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 
+@OptIn(ExperimentalCoroutinesApi::class)
 class WorkerBinderTest {
+
+  @get:Rule val ribCoroutinesRule = RibCoroutinesRule()
+
   private val worker: Worker = mock()
+  private val workerBinderListener: WorkerBinderListener = mock()
+
+  private val fakeWorker = FakeWorker()
+  private val interactor = FakeInteractor<Presenter, Router<*>>()
+
+  @Before
+  fun setUp() {
+    WorkerBinder.initializeMonitoring(workerBinderListener)
+  }
 
   @Test
   fun bind_whenInteractorAttached_shouldStartWorker() {
@@ -146,6 +165,43 @@ class WorkerBinderTest {
     verify(worker, times(1)).onStop()
     unbinder.unbind()
     verify(worker, times(1)).onStop()
+  }
+
+  @Test
+  fun bind_withUnconfinedCoroutineDispatchers_shouldReportBinderInformationForOnStart() = runTest {
+    val binderDurationCaptor = argumentCaptor<WorkerBinderInfo>()
+    bindFakeWorker()
+    verify(workerBinderListener).onBindCompleted(binderDurationCaptor.capture())
+    binderDurationCaptor.firstValue.assertWorkerDuration(
+      "FakeWorker", WorkerEvent.START, RibDispatchers.Unconfined
+    )
+  }
+
+  @Test
+  fun unbind_withUnconfinedCoroutineDispatchers_shouldReportBinderDurationForOnStop() = runTest {
+    val binderDurationCaptor = argumentCaptor<WorkerBinderInfo>()
+    val unbinder = bindFakeWorker()
+    unbinder.unbind()
+    verify(workerBinderListener, times(2)).onBindCompleted(binderDurationCaptor.capture())
+    binderDurationCaptor.secondValue.assertWorkerDuration(
+      "FakeWorker", WorkerEvent.STOP, RibDispatchers.Unconfined
+    )
+  }
+
+  private fun bindFakeWorker(): WorkerUnbinder {
+    interactor.attach()
+    interactor.enableTestScopeOverride()
+    return bind(interactor, fakeWorker)
+  }
+
+  private fun WorkerBinderInfo.assertWorkerDuration(
+    expectedWorkerClassName: String,
+    expectedWorkerEvent: WorkerEvent,
+    expectedWorkerBinderThreadingType: CoroutineDispatcher
+  ) {
+    assertThat(workerName).contains(expectedWorkerClassName)
+    assertThat(workerEvent).isEqualTo(expectedWorkerEvent)
+    assertThat(coroutineDispatcher).isEqualTo(expectedWorkerBinderThreadingType)
   }
 }
 
--- a/android/libraries/rib-base/build.gradle
+++ b/android/libraries/rib-base/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     compileOnly deps.apt.androidApi
     compileOnly deps.external.checkerQual
 
+    testImplementation project(":libraries:rib-coroutines-test")
     testImplementation deps.androidx.annotations
     testImplementation deps.apt.androidApi
     testImplementation deps.kotlin.stdlib
--- a/android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/WorkerTest.kt
+++ b/android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/WorkerTest.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2023. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.core
+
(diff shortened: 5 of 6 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerMonitoringListener.kt
```diff
@@ -0,0 +1,9 @@
+package com.ubercab.presidio.plugin.rib_worker_binder
+
+interface RibWorkerMonitoringListener {
+    fun onWorkerStartCompleted(
```
FranAguilera, line 4: This will be renamed to drop the onStart

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerThreadingType.kt
```diff
@@ -0,0 +1,8 @@
+package com.uber.rib.core
+
+enum class RibWorkerThreadingType {
+    BACKGROUND_COMPUTATION,
+    BACKGROUND_IO,
+    MAIN,
+    CALLER_THREAD
+}
```
FranAguilera, line 8: I'm thinking of adding a custom executor with one single thread as well

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -75,22 +224,26 @@ object WorkerBinder {
...
-      .takeUntil { workerEvent: WorkerEvent -> workerEvent === WorkerEvent.STOP }
+    val unbindFlow = MutableSharedFlow<WorkerEvent>(0, 1, BufferOverflow.DROP_OLDEST)
+
+    val workerLifecycle = merge(mappedLifecycle.asFlow(), unbindFlow)
+      .transformWhile { emit(it) ; it != WorkerEvent.STOP }
+      .asObservable()
+
     bindToWorkerLifecycle(workerLifecycle, worker)
-    return WorkerUnbinder { unbindSubject.accept(WorkerEvent.STOP) }
+    return WorkerUnbinder { unbindFlow.tryEmit(WorkerEvent.STOP) }
```
FranAguilera, line 246: looks like I was behind. Will update later on

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -21,15 +21,164 @@ import com.uber.rib.core.lifecycle.InteractorEvent
...
+
+  @JvmStatic
+  private var ribWorkerBindMonitoringListeners: MutableSet<RibWorkerBindMonitoringListener> = mutableSetOf()
+
+  /**
+   * Adds [RibWorkerBindMonitoringListener]
+   */
+  @JvmStatic
+  fun addMonitoringListener(ribWorkerBindMonitoringListener: RibWorkerBindMonitoringListener) {
+    this.ribWorkerBindMonitoringListeners.add(ribWorkerBindMonitoringListener)
```
FranAguilera, line 46: will update this

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerBindMonitoringListener.kt
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.core
+
+/**
+ * Reports total Worker Binding duration [RibWorkerThreadingType]
+ */
+interface RibWorkerBindMonitoringListener {
```
psteiger, line 21: Since it's a single method interface, we could make it a `fun interface` for idiomatic instantiation

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerBindMonitoringListener.kt
psteiger, line 1: I would not create a new file for this, this interface could be on `WorkerBinder.kt`

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerThreadingType.kt
psteiger, line 1: Likewise, I would not create a new file for this, this interface could be on `WorkerBinder.kt`

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -21,15 +21,164 @@ import com.uber.rib.core.lifecycle.InteractorEvent
...
+   *
+   * Worker.onStart/onStop will run on the specified [RibWorkerThreadingType]
+   *
+   * @param worker The class that wants to be informed when to start and stop doing work.
+   * @param workerLifecycle Observable<WorkerEvent>
+   * @param ribWorkerThreadingType Rib Threading type where Worker.onStart/onStop will operate
+   */
+  @JvmStatic
+  fun bind(
+    workerToBeBound: Worker,
```
psteiger, line 61: I'd just call the parameter `worker` instead, it's obvious it's the worker to be bound.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerBindMonitoringListener.kt
FranAguilera: Thanks, let me update in a bit

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibWorkerBindMonitoringListener.kt
FranAguilera: Great suggestions. Updated

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: Totally agree, updated as well

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -155,10 +294,42 @@ object WorkerBinder {
     worker: Worker
   ) {
     workerLifecycle.subscribe { workerEvent: WorkerEvent ->
+
       when (workerEvent) {
-        WorkerEvent.START -> worker.onStart(WorkerScopeProvider(workerLifecycle.hide()))
-        else -> worker.onStop()
+        WorkerEvent.START -> {
+          val onStartTotalDurationMilli = measureTimeMillis {
+            worker.onStart(WorkerScopeProvider(workerLifecycle.hide()))
```
FranAguilera, line 301: Adding monitoring now will allow to monitor long Worker.onStart durations happening in main thread in current WorkerBinder.bind usages

FranAguilera:
- To rebase when changes of https://github.com/uber/RIBs/pull/552 are landed
- To add threading type again for Java support
- To revisit monitoring listener setup

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -21,15 +21,164 @@ import com.uber.rib.core.lifecycle.InteractorEvent
...
+import kotlin.system.measureTimeMillis
 
 /** Helper class to bind to an interactor's lifecycle to translate it to a [Worker] lifecycle. */
 object WorkerBinder {
+
+  /**
+   * A set of [RibWorkerBinderDurationListener]
+   */
+  @JvmStatic
+  val ribWorkerBinderDurationListeners: MutableSet<RibWorkerBinderDurationListener> = mutableSetOf()
```
FranAguilera, line 42: Note for reviewer: Will update this monitoring setup soon

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -21,15 +21,164 @@ import com.uber.rib.core.lifecycle.InteractorEvent
...
+   *
+   * @param worker The Worker to be bound
+   * @param interactor Interactor
+   * @param threadingType ThreadingType. Threading type where worker will operate.
+   */
+  @JvmStatic
+  fun bind(
+    worker: Worker,
+    interactor: Interactor<*, *>,
+    threadingType: ThreadingType
```
FranAguilera, line 93: Still debating if we should do this other than caller thread. Given the existing binding led to some misconceptions where Workers will be running upon. Maybe we make this new binders be defaulted to background

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -33,4 +35,63 @@ interface Worker {
...
+
+/**
+ * Reports Worker total binding duration details
+ */
+interface RibWorkerBinderDurationListener {
+
+  /**
+   * Reports when Worker.onStart/Worker.onStop has completed
+   */
+  fun onBindCompleted(
```
psteiger, line 94: Still feels a bit random to me to have duration metrics on the API itself.

Also the fact that this would apply only to `Worker` is unconvicing. I mean, why are we reporting `Worker` duration on callback but not `Interactor` `didBecomeActive` duration on attach by router? It feels inconsistent. 

Have you given a thought about wrapping `Worker` instead? 

```kotlin
class MonitoredWorker(val worker: Worker) : Worker {
  fun onStartCompleted(duration: Long) {...}
  fun onStopCompleted(duration: Long) {...}
  
  override fun onStart(lifecycle: WorkerScopeProvider) {
    val duration = measureTimeMillis { worker.onStart(lifecycle) }
    onStartCompleted(time)
  }
  
  override fun onStop() {
    val duration = measureTimeMillis { worker.onStop() }
    onStopCompleted(time)
  }
}

File: android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt [hunk 11]
```diff
@@ -33,4 +35,63 @@ interface Worker {
   /** Called when the worker is stopped.  */
   @JvmDefault
   fun onStop() {}
+
+  /**
+   * Allows overriding the threading type where the Worker.onStart/Worker.onStop will be called upon
+   */
+  @JvmDefault
+  fun threadingType(): ThreadingType = ThreadingType.CALLER_THREAD
```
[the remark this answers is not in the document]
FranAguilera: nice! will update

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
FranAguilera: I agree on the lacking monitoring for Interactor's (will look into another PR for this).

The main reason to keep this coupled to Worker are:

- Would be great to have a complete view of duration of all Workers without changing current Worker usages. Wrapping is a possibility but that will require major refactor in order to get this Monitored Worker. 

-  Still this new monitoring is really tied to ThreadingType which I would like to pass on the callback as well with the thread name, worker class name (I know this will be obfuscated on release builds but still useful info)

psteiger: For this point
> Wrapping is a possibility but that will require major refactor in order to get this Monitored Worker.

Not really, since all would be needed was to call 

`worker.asMonitoredWorker()`

At binding site 

```
WorkerBinder.bind(worker.asMonitoredWorker())

fun Worker.asMonitoredWorker() = MonitoredWorker(this)
```

psteiger: For this point
> Wrapping is a possibility but that will require major refactor in order to get this Monitored Worker.

Not really, since all would be needed was to call 

`worker.asMonitoredWorker()`

At binding site 

```
WorkerBinder.bind(worker.asMonitoredWorker())

fun Worker.asMonitoredWorker() = MonitoredWorker(this)
```

FranAguilera: Still there could be cases where a Worker implementation extends other classes so this may not be suitable for all. 

Also the consumer of WorkerBinder will need to explicit opt in for this monitoring, so this will defer with the main goal of global a complete monitoring of all existing workers running

FranAguilera:
Will remove threading type from WorkerBinder.bind apis

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -21,15 +21,173 @@ import com.uber.rib.core.lifecycle.InteractorEvent
...
+   * NOTE: If worker overrides ThreadingType other than ThreadingType.CALLER_THREAD. Worker threading
+   * type will take priority over the one specified at binding level
+   *
+   * @param worker The class that wants to be informed when to start and stop doing work.
+   * @param workerLifecycle Observable<WorkerEvent>
+   * @param coroutineScope coroutineScope
+   * @param threadingType ThreadingType. Threading type where worker will operate.
+   */
+  @JvmStatic
+  fun bind(
```
FranAguilera, line 67: These new methods will be removed and update just this internals

FranAguilera:
Will fix this bad rebase later on

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -89,6 +108,24 @@ public object WorkerBinder {
 
   @JvmStatic
   @VisibleForTesting
+  public fun bind(mappedLifecycle: Observable<WorkerEvent>, coroutineScope: CoroutineScope, worker: Worker): WorkerUnbinder {
+    val unbindFlow = MutableSharedFlow<WorkerEvent>(0, 1, BufferOverflow.DROP_OLDEST)
```
FranAguilera, line 112: Will update after https://github.com/uber/RIBs/pull/552 is landed

FranAguilera:
Will be updating after rebasing with https://github.com/uber/RIBs/pull/552

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -164,25 +187,107 @@ public object WorkerBinder {
...
+    val coroutineStart = if (threadingType == ThreadingType.WORKER_BINDER_CALLER_THREAD) {
+      CoroutineStart.UNDISPATCHED
+    } else {
+      CoroutineStart.DEFAULT
+    }
+    /*
+     * GlobalScope won't leak the job, because the flow completes when lifecycle completes.
+     */
+    @OptIn(DelicateCoroutinesApi::class)
+    val job = GlobalScope.launch(coroutineDispatcher, start = coroutineStart) {
```
FranAguilera, line 206: Moving up in order to allow easier access to duration listener

android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/WorkerBinderTest.kt
```diff
@@ -24,14 +24,22 @@ import com.uber.rib.core.WorkerBinder.mapPresenterLifecycleToWorker
...
 import com.uber.rib.core.lifecycle.WorkerEvent
+import org.junit.Before
 import org.junit.Test
 import org.mockito.kotlin.any
+import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
-
 class WorkerBinderTest {
```
FranAguilera, line 34: Will add more test

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -22,6 +22,16 @@ package com.uber.rib.core
  * other lifecycles we're interested in.
  */
 public interface Worker {
+
+  /**
+   * Specifies in which ThreadingType the WorkerBinder will bound this Worker.
+   *
+   * Whe not overriden, will use the thread where [WorkerBinder.bind] call happens
```
FranAguilera, line 29: Will fix these typos

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -88,6 +102,7 @@ public object WorkerBinder {
 
   @JvmStatic
   @VisibleForTesting
+  @Deprecated("Use WorkerBinder.bind(lifecycleFlow,lifecycleRange)")
```
psteiger, line 105: Good call. I propose we add a `ReplaceWith()` alongside the message

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -22,6 +22,16 @@ package com.uber.rib.core
...
+   */
+  @JvmDefault
+  public val workerBinderThreadingType: WorkerBinderThreadingType
+    get() = WorkerBinderThreadingType.CALLER_THREAD
```
psteiger, line 33: Decided against defaulting to `BACKGROUND_COMPUTATION` for retro compatibility?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -178,11 +272,54 @@ private fun <T : Comparable<T>> Worker.bind(
...
+      .collect {
+        val workerScopeProvider = WorkerScopeProvider(lifecycle.asScopeProvider(lifecycleRange))
+        val totalBindingDurationMilli = measureTimeMillis { onStart(workerScopeProvider) }
+
+        workerBinderDurationListeners.reportDuration(
+          this@bind,
+          WorkerEvent.START,
+          totalBindingDurationMilli
+        )
+      }
```
psteiger, line 299: For simplicity we could make an extension like:
```kotlin
private inline fun Worker.reportDuration(event: WorkerEvent, block: Worker.() -> Unit) {
  val duration = measureTimeMillis { block() }
  workerBinderDurationListeners.reportDuration(
      this@bind,
      WorkerEvent.STOP,
      totalBindingDurationMilli
   )
}
```
And we can use like:
```kotlin
    lifecycle
      .takeWhile { it < lifecycleRange.endInclusive }
      .onCompletion { reportDuration(WorkerEvent.STOP) { onStop() } }
      .collect { 
        reportDuration(WorkerEvent.START) { 
          onStart(WorkerScopeProvider(lifecycle.asScopeProvider(lifecycleRange)))
        }
      }
```

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -21,16 +21,30 @@ import com.uber.rib.core.lifecycle.InteractorEvent
...
+
+  private var workerBinderDurationListeners: MutableSet<WorkerBinderDurationListener> = mutableSetOf()
+
+  /**
+   * Adds [WorkerBinderDurationListener] in order to report Worker.onStart/onStop duration info.
+   *
+   */
+  @JvmStatic
+  public fun addDurationListener(workerBinderDurationListener: WorkerBinderDurationListener) {
+    this.workerBinderDurationListeners.add(workerBinderDurationListener)
```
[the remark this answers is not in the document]
FranAguilera: So far on the internal monorepo verification couldn't see any leak reports for this but will double check again with profiler

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -166,10 +181,89 @@ public object WorkerBinder {
...
+private data class CoroutineScopeLauncher(
+  val coroutineDispatcher: CoroutineDispatcher,
+  val coroutineStart: CoroutineStart
+)
+
+private fun mapToCoroutineScopeLauncher(workerBinderThreadingType: WorkerBinderThreadingType): CoroutineScopeLauncher {
+  val coroutineDispatcher = when (workerBinderThreadingType) {
+    WorkerBinderThreadingType.BACKGROUND_COMPUTATION -> RibDispatchers.Default
+    WorkerBinderThreadingType.BACKGROUND_IO -> RibDispatchers.IO
+    WorkerBinderThreadingType.MAIN -> RibDispatchers.Main
```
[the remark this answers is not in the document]
FranAguilera: got it, will update

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
FranAguilera: That's right. Given this will be running on existing usages we need to guarantee the current threading model is honored (otherwise we may face unwanted concurrent/not on ui thread crashes)

With the added monitoring will have an option now to detect workers on UI thread and move those to background

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: thanks! this looks so much cleaner. Will apply soon

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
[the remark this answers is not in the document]
FranAguilera: Ended adding RibDispatchers.Main.immediate to prevent confusion with Java usages

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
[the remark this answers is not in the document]
FranAguilera: Couldn't find any leak after profiling, but was thinking to use WeakReference for the stored listeners just in case

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -178,11 +284,51 @@ private fun <T : Comparable<T>> Worker.bind(
...
   return WorkerUnbinder(job::cancel)
 }
+
+private inline fun Worker.bindAndReportDuration(
+  workerBinderDurationListeners: WeakReference<WorkerBinderDurationListener>?,
+  event: WorkerEvent,
+  workerBindAction: Worker.() -> Unit
+) {
+  val duration = measureTimeMillis { workerBindAction() }
+  workerBinderDurationListeners.reportDuration(this, event, duration)
```
tyvsmith, line 312: Since this was nullable and weak ref, it needs to be checked before calling the method, right?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
tyvsmith, line 33: Assuming we wanted to default to background to prevent any new additions from being made unless the explicitly indicated calling thread, what would that look like and how could we migrate internally (ie, new internal base class with a linter against it and baseline it)? Anything reasonable that doesn't result in a massive migration?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -166,10 +193,89 @@ public object WorkerBinder {
...
+
+  /**
+   * Worker event type (START/STOP)
+   */
+  val workerEvent: WorkerEvent,
+
+  /**
+   * Threading type where Worker.onStart/onStop upon calling [WorkerBinder.bind]
+   */
+  val workerBinderThreadingType: WorkerBinderThreadingType,
```
tyvsmith, line 223: Since the data doesn't actually tell us which thread, just if it was calling thread of parent, do we want to track that here, or will be in a custom impl internally?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -178,11 +284,51 @@ private fun <T : Comparable<T>> Worker.bind(
...
+private inline fun Worker.bindAndReportDuration(
+  workerBinderDurationListeners: WeakReference<WorkerBinderDurationListener>?,
+  event: WorkerEvent,
+  workerBindAction: Worker.() -> Unit
+) {
+  val duration = measureTimeMillis { workerBindAction() }
+  workerBinderDurationListeners.reportDuration(this, event, duration)
+}
+
+private fun WeakReference<WorkerBinderDurationListener>?.reportDuration(
```
tyvsmith, line 315: Does an extension function on a nullable run properly if the value is null, or only in cases when it's satisfied as nonnull?

FranAguilera: Looks like it will run as satisfied as non-null. Still will update for clarity as non-null and update the caller

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: Was thinking to use `threadName` within the `WorkerBinderDuration` data class to be called on each onStart/onStop completion

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: It seems to be the extension function will trigger only when non nullable, but will update the call per clarity

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
FranAguilera: Was thinking we could rely on the new listener to trigger non-fatals for expensive UI workers (e.g. crash on debug builds) and have global monitoring that will report expensive UI workers. 

Given that @psteiger is proposing a new RibCoroutineWorker (in https://github.com/uber/RIBs/pull/547)  that will run by default on the background, internally we could recommend to use this new definition (with suspend function), mark as deprecated this existing old definition and keep it for backward compatibility with existing usages

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -178,11 +286,51 @@ private fun <T : Comparable<T>> Worker.bind(
...
+private inline fun Worker.bindAndReportDuration(
+  workerBinderDurationListeners: WeakReference<WorkerBinderDurationListener>?,
+  event: WorkerEvent,
+  workerBindAction: Worker.() -> Unit
+) {
+  val duration = measureTimeMillis { workerBindAction() }
+  workerBinderDurationListeners?.reportDuration(this, event, duration)
+}
+
+private fun WeakReference<WorkerBinderDurationListener>.reportDuration(
```
FranAguilera, line 317: @tyvsmith Updated definition as non-nullable for clarification

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -166,10 +193,62 @@ public object WorkerBinder {
   }
 }
 
+/**
+ * Worker.onStart/onStop duration information
+ */
+public data class WorkerBinderDuration(
```
tyvsmith, line 199: The naming "WorkerBinderDuration" feels a bit too implementation specific on duration. This is more of a worker metadata + callback. What do you think about naming it more generally?

FranAguilera: Yes, it makes sense. Will update in a bit

tyvsmith: If we like this pattern after rollout, maybe it makes sense to have an Interactor Metadata class too.

FranAguilera: @tyvsmith Updated the naming

tyvsmith (approved):
Looks good, thanks for the update.

PR #558, opened 2023-04-23
Upgrade code formatters versions

SergeySmykovskyi:
**Description**:

Updated formatters versions:
- gjf `1.7` -> `1.16.0`
- ktfmt `0.23` -> `0.43`
- ktlint `0.41.0` -> `0.48.2`
- spotless `5.11.0` -> `6.18.0`

Most changes done by `spotless` but upgrading `ktlint` version produced "Incorrect package name" error. That's why few packages which used underscores was renamed to the camel case style [Coding conventions / Naming rules
](https://kotlinlang.org/docs/coding-conventions.html#naming-rules)

Manual changes :
- Two gradle warnings auto fixes by IDE : 
  - `tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {` -> `tasks.withType(KotlinCompile).configureEach {`
  - `task clean(type: Delete) {` -> `tasks.register('clean', Delete) {`
-  Enabled `ktfmt` with `googleStyle()` ([android/build.gradle](https://github.com/uber/RIBs/compare/main...SergeySmykovskyi:update_ktfmt?expand=1#diff-197b190e4a3512994d2cebed8aff5479ff88e136b8cc7a4b148ec9c3945bd65a))
- Replaced `userData` by `editorConfigOverride`  for `ktlint` configuration ([android/build.gradle](https://github.com/uber/RIBs/compare/main...SergeySmykovskyi:update_ktfmt?expand=1#diff-197b190e4a3512994d2cebed8aff5479ff88e136b8cc7a4b148ec9c3945bd65a))
- Renamed `tutorial` and `demos` inner packages
  - Renamed `logged_in` package -> `loggedin`
  - Renamed `logged_out` package -> `loggedout`
  - Renamed `off_game` package -> `offgame`
  - Renamed `tic_tac_toe` package -> `tictactoe`

**Test/verification**
No manual or unit testing required, code format changes only, if `CI` pass everything should be fine


psteiger:
I concur that using snake case is bad for package name, but IMO camel case is even worse. If we want to get rid of snake case, I'd prefer:
logged_in -> loggedin
logged_out -> loggedout
off_game -> offgame
tic_tac_toe -> tictactoe

SergeySmykovskyi:
> I concur that using snake case is bad for package name, but IMO camel case is even worse. If we want to get rid of snake case, I'd prefer: logged_in -> loggedin logged_out -> loggedout off_game -> offgame tic_tac_toe -> tictactoe

Don't have personal preference on package naming but the camel case looked more readable. I think simple concatenation still is prefered option for libraries, so make sense to rename by your recommendation 👍

SergeySmykovskyi:
> What's the current gap between ktfmt and ktlint for auto formatting? Can/should we use one exclusively (ktfmt?). Where's other popular OSS libraries heading on this one?

After quick research in most common kotlin android libraries - popular options is `ktlint` or `detekt`(flexible wrapper over `ktlint`). If we need to switch single one, IMO better to go with `ktlint` (`detekt` not available by spotless plugin) `ktlint_official` code style ([Release 0.49.0](https://github.com/pinterest/ktlint/releases/tag/0.49.0), haven't chance to bump version to the 0.49.0 as it was release only few days ago + might need additional changes). The `ktfmt` only formatting code based on `gjf` and don't check kotlin coding conventions.

tyvsmith:
> > What's the current gap between ktfmt and ktlint for auto formatting? Can/should we use one exclusively (ktfmt?). Where's other popular OSS libraries heading on this one?
> 
> After quick research in most common kotlin android libraries - popular options is `ktlint` or `detekt`(flexible wrapper over `ktlint`). If we need to switch single one, IMO better to go with `ktlint` (`detekt` not available by spotless plugin) `ktlint_official` code style ([Release 0.49.0](https://github.com/pinterest/ktlint/releases/tag/0.49.0), haven't chance to bump version to the 0.49.0 as it was release only few days ago + might need additional changes). The `ktfmt` only formatting code based on `gjf` and don't check kotlin coding conventions.

Detekt's autoformatting just uses ktlint under the hood, so it's really just a discussion of ktlint vs ktfmt.

We use both internally as well, and that's because at the time of integration ktlint didn't support things like line breaks, but ktfmt had some weird things it did and ktlint would clean up. GJF is a long time stable formatter, so fundamentally ktfmt probably makes more sense. I'm just not sure if it's solved all the problems at this point, or if we still need both. Or if ktfmt has taken on those gaps and the industry has aligned. For now, we don't need to decide, but it's something to keep an eye on. CC @jbarr21 in case you have a current opinion.

tyvsmith:
I'm not a fan of camelcase packages, let's do what @psteiger suggested.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/RootView.kt
```diff
@@ -33,7 +35,7 @@ class RootView @JvmOverloads constructor(
       TextView(context).apply {
         text = "root (view)"
         setTextColor(Color.WHITE)
-      }
+      },
```
psteiger, line 38: I don't get this trailing comma on here. (I do get and support the trailing comma on constructor). Is this auto by `googleStyle`?

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/RootView.kt
```diff
@@ -21,10 +21,12 @@ import android.util.AttributeSet
 import android.widget.FrameLayout
 import android.widget.TextView
 
-class RootView @JvmOverloads constructor(
+class RootView
+@JvmOverloads
+constructor(
```
psteiger, line 26: Previous form was superior IMO

SergeySmykovskyi: This change was done by `ktfmt`, we cannot customize this behavior but we can keep `ktfmt` disabled for now.

android/demos/compose/src/main/kotlin/com/uber/rib/compose/root/RootView.kt
SergeySmykovskyi: That's auto by `ktlint` since [Trailing comma on call site](https://pinterest.github.io/ktlint/rules/standard/#trailing-comma-on-call-site) became one of standard rules

tyvsmith (approved):

PR #563, opened 2023-04-27
Use Kotlin contracts to remove `var` and `!!` usage in `RibCoroutineWorker`

psteiger:
Coroutines started with `CoroutineStart.UNDISPATCHED` is guaranteed to run its block at least until the first check for cancellation / failure point, even when `CoroutineScope` is already cancelled, so we can use Kotlin contracts to have a non-nullable `bindJob`, and a `val` instead of `var`.

Incidentally, applies new spotless since when `RibCoroutineWork` was merged, spotless checks were already done when a new ktlint version/config was merged.


FranAguilera (approved):

tyvsmith (approved):
Looks good.

PR #564, opened 2023-04-27
Provide option to bind multiple Workers at once on specific RibDispatchers

FranAguilera:
## Goal

Given the need to change the dispatcher of multiple workers being bound at once, updating WorkerBinder to support this.

Following great suggestion from @psteiger , will also update the Worker contract to have a default `coroutineDispatcher`as null. When the `Worker`'s `coroutineDispatcher` is overriden with a valid dispatcher, the resulting dispatcher will always be applied instead of the one passed in the binding methods. This is really similar to RibCoroutineDispatcher as well

**NOTE: By default WorkerBinder dispatcher remains as Unconfined to keep the existing threading model of workers**

## Sample

Given 

`class FakeWorker: Worker
`

`class FeatureWorker:Worker
`

```
class UiWorker : Worker {
  override val coroutineDispatchers: CoroutineDispatchers? = RibDispatchers.Main
}
```
And a list of workers to be bound

`val workers = listOf(fakeWorker, featureWorker, uiWorker)
`
Upon calling:

`WorkerBinder.bind(interactor, workers, RibDispatchers.Default)
`

Results in:

`FakeWorker` and `FeatureWorker` will be bound on `RibDispatchers.Default` and `UiWorker` will always be bound on `RibDispatchers.Main`

## Related issue(s)

https://github.com/uber/RIBs/issues/565

## Verification

- Verify on demo project by adding custom workers (will add sample app with Workers on following PR)
- Verify locally on internal monorepo

### DIFF
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
@@ -192,13 +192,13 @@ public fun Worker.asRibCoroutineWorker(): RibCoroutineWorker =
 /** Converts a [RibCoroutineWorker] to a [Worker]. */
 @JvmOverloads
 public fun RibCoroutineWorker.asWorker(
-  dispatcher: CoroutineDispatcher? = null,
-): Worker = RibCoroutineWorkerToWorkerAdapter(this, dispatcher)
+  coroutineContext: CoroutineContext = EmptyCoroutineContext,
+): Worker = RibCoroutineWorkerToWorkerAdapter(this, coroutineContext)
 
 internal open class WorkerToRibCoroutineWorkerAdapter(private val worker: Worker) :
   RibCoroutineWorker {
   override suspend fun onStart(scope: CoroutineScope) {
-    withContext(worker.coroutineDispatcher) { worker.onStart(scope.asWorkerScopeProvider()) }
+    withContext(worker.coroutineContext) { worker.onStart(scope.asWorkerScopeProvider()) }
   }
 
   override fun onStop(cause: Throwable): Unit = worker.onStop()
@@ -206,17 +206,17 @@ internal open class WorkerToRibCoroutineWorkerAdapter(private val worker: Worker
 
 internal open class RibCoroutineWorkerToWorkerAdapter(
   private val ribCoroutineWorker: RibCoroutineWorker,
-  private val dispatcher: CoroutineDispatcher?,
+  coroutineContext: CoroutineContext,
 ) : Worker {
-  override val coroutineDispatcher: CoroutineDispatcher
-    get() = dispatcher ?: super.coroutineDispatcher
+
+  override val coroutineContext: CoroutineContext = coroutineContext + super.coroutineContext
 
   override fun onStart(lifecycle: WorkerScopeProvider) {
     // We can start it undispatched because Worker binder will already call `onStart` in correct
     // context,
     // but we still want to pass in `coroutineDispatcher` to resume from suspensions in `onStart` in
     // correct context.
-    lifecycle.coroutineScope.launch(coroutineDispatcher, start = CoroutineStart.UNDISPATCHED) {
+    lifecycle.coroutineScope.launch(coroutineContext, start = CoroutineStart.UNDISPATCHED) {
       supervisorScope { ribCoroutineWorker.onStart(this) }
     }
   }
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
@@ -204,12 +204,11 @@ internal open class WorkerToRibCoroutineWorkerAdapter(private val worker: Worker
   override fun onStop(cause: Throwable): Unit = worker.onStop()
 }
 
-internal open class RibCoroutineWorkerToWorkerAdapter(
+internal open class RibCoroutineWorkerToWorkerAdapter
+internal constructor(
   private val ribCoroutineWorker: RibCoroutineWorker,
-  coroutineContext: CoroutineContext,
+  override val coroutineContext: CoroutineContext,
 ) : Worker {
-
-  override val coroutineContext: CoroutineContext = coroutineContext + super.coroutineContext
 
   override fun onStart(lifecycle: WorkerScopeProvider) {
     // We can start it undispatched because Worker binder will already call `onStart` in correct
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
@@ -15,6 +15,8 @@
  */
 package com.uber.rib.core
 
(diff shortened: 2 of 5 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -194,6 +236,19 @@ public object WorkerBinder {
   }
 }
 
+/**
+ * Holds relevant information for [dispatcherAtBinder] [workerCoroutineWorker] and resulting
+ * [jobCoroutineWorker]
+ */
+public data class CoroutineContextInfo(
```
tyvsmith, line 243: This entire thing need to be public?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -194,6 +236,19 @@ public object WorkerBinder {
...
+  /** Dispatcher passed at WorkerBinder.bind call */
+  val dispatcherAtBinder: CoroutineDispatcher,
+  /** CoroutineContext defined at Worker */
+  val workerCoroutineWorker: CoroutineContext,
+  /** Effective CoroutineContext that will be used by the Worker binding */
+  val jobCoroutineWorker: CoroutineContext,
```
tyvsmith, line 249: Anything actually need to be held onto besides the jobCoroutineWorker? Feels like this may be confusing for consumers to understand the nuance of.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -15,25 +15,37 @@
...
+   * 1) workers = listOf(defaultWorker, uiWorker)
+   * 2) Where [uiWorker] overrides [coroutinesContext] with [RibDispatchers.Main]
+   * 3) After calling WorkerBinder.bind(interactor, workers, RibDispatchers.IO). [uiWorker] will be
+   *    guaranteed to be called on RibDispatchers.Main
    */
   @JvmDefault
-  public val coroutineDispatcher: CoroutineDispatcher
-    get() = RibDispatchers.Unconfined
+  public val coroutineContext: CoroutineContext
+    get() = EmptyCoroutineContext
```
tyvsmith, line 48: Makes sense.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: Good catch! definitely not needed

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: I agree, let's keep just the jobCoroutineWorker then

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
FranAguilera: Ended defaulting with nullable `coroutineDispatcher` for supporting override on existing Workers defined on Java

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -15,25 +15,35 @@
...
+   * 1) workers = listOf(defaultWorker, uiWorker)
+   * 2) Where [uiWorker] overrides [coroutineDispatcher] with [RibDispatchers.Main]
+   * 3) After calling WorkerBinder.bind(interactor, workers, RibDispatchers.IO). [uiWorker] will be
+   *    guaranteed to be called on RibDispatchers.Main
    */
   @JvmDefault
-  public val coroutineDispatcher: CoroutineDispatcher
-    get() = RibDispatchers.Unconfined
+  public val coroutineDispatcher: CoroutineDispatcher?
+    get() = null
```
FranAguilera, line 46: @psteiger Ended choosing a nullable dispatcher instead of EmptyCoroutineContext for easier Java support on existing Workers

psteiger: Out of curiosity, what was the issue with Java + EmptyCoroutineContext?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -198,25 +198,27 @@ public fun RibCoroutineWorker.asWorker(
...
+    withContext(dispatcher) { worker.onStart(scope.asWorkerScopeProvider()) }
   }
 
   override fun onStop(cause: Throwable): Unit = worker.onStop()
 }
 
 internal open class RibCoroutineWorkerToWorkerAdapter(
   private val ribCoroutineWorker: RibCoroutineWorker,
-  private val dispatcher: CoroutineDispatcher?,
+  dispatcher: CoroutineDispatcher?,
```
psteiger, line 210: We can just use `override val coroutineDispatcher` in constructor instead

FranAguilera: Nice catch, will update

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
FranAguilera: I was getting an impossible cast from Dispatchers to CoroutineContext on java files. Let me double check the hierarchy to see what's going on

FranAguilera: Please ignore this, I was mistaken during local monorepo bump. This should always be a possible cast following. Will revert back

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -208,15 +208,16 @@ internal open class RibCoroutineWorkerToWorkerAdapter(
   private val ribCoroutineWorker: RibCoroutineWorker,
   private val dispatcher: CoroutineDispatcher?,
```
psteiger, line 209: Now that we're using `CoroutineContext` in `Worker`, we can receive a `CoroutineContext` in here instead, and do:
```
  override val coroutineContext: CoroutineContext = coroutineContext + super.coroutineContext
```

`super.coroutineContext` will override elements already present in `this.coroutineContext` (if any)

FranAguilera: good catch. Will update

tyvsmith (approved):
Looks good, thanks. Please wait for @psteiger to confirm his comments as well.

psteiger (approved):
Looks good! Thanks, I'm pretty happy with Worker using CoroutineContext for controlling execution context, that's a clean solution, and interop with RibCoroutineWorker is now much more natural.

FranAguilera:
@psteiger @tyvsmith  Thank you folks! Landing now

PR #567, opened 2023-04-28
Bump Kotlin, Gradle, and other dependencies versions.

### LINKED ISSUE #566: Update Gradle and AGP to 8.x

psteiger:
Fixes #566 

Kotlin 1.7.10 -> 1.8.20
Gradle 7.3.3 -> 8.1.1
AndroidX Lifecycle ViewModel Compose 1.0.0-alpha05 -> 2.5.1
AndroidX Activity 1.3.0-alpha08 -> 1.7.0
AndroidX Activity KTX 1.3.0-beta02 -> 1.7.0
AndroidX Activity Compose 1.3.0-beta02 -> 1.7.0
Compose compiler 1.3.0 -> 1.4.6
Compose libraries 1.2.1 -> 1.4.0
Java version 1.8 -> 11
IntelliJ plugin 1.11.0 -> 1.13.1 (1.13.2 onwards fails test `ribGenerators_shouldGenerateClassesThatCompiler` for reasons beyond my understanding, cc @idanakav for vis)
AGP 7.2.2 -> 7.4.2 (8.0 not supported in IJ 2023.1 bundled Android plugin)
gradle-maven-publish 0.18.0 -> 0.25.2

### Additional changes

- Move package string from manifests to `android.namespace` in Gradle build scripts.
- Remove `targetSdk` from Android libraries (only applications should provide it)
- Remove `buildToolsVersion`.
- `compileSdkVersion`, `minSdkVersion`, `targetSdkVersion` swapped to `compileSdk`, `minSdk`, `targetSdk`.
- Disablement of `buildConfig` is now done by `buildFeatures.buildConfig = false` instead of in the project `build.gradle`.
- Added `exported="true"` in demo Android projects activities.
- `-Xfriend-paths` `kotlinc` option seemingly only works now if we use `@file:Suppress("invisible_reference", "invisible_member")` in files that access internal components.
- `androidx.lifecycle.ViewTreeLifecycleOwner(View, LifecycleOwner)` has been replaced with `View.setViewTreeLifecycleOwner(LifecycleOwner)`.
- Removed unnecessary dependencies on Kotlin stdlib, as the Kotlin JVM plugin automatically adds it. For reasons beyond my understanding, removing it from `android/libraries/rib-test` fails gradle sync with the following error, so it's the only place it's still kept.
```
com.intellij.openapi.externalSystem.model.ExternalSystemException: Could not resolve all dependencies for configuration ':libraries:rib-coroutines:testCompileClasspath'.
Cannot change dependencies of dependency configuration ':libraries:rib-test:api' after it has been included in dependency resolution.
```

### Tests

Did a quick smoke test with a `-LOCAL` build against our monorepo.

### DIFF
--- a/android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
+++ b/android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("invisible_reference", "invisible_member")
+
 package com.uber.rib.core
 
 import android.app.Application
--- a/android/libraries/rib-workflow-test/src/main/AndroidManifest.xml
+++ b/android/libraries/rib-workflow-test/src/main/AndroidManifest.xml
@@ -1,2 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<manifest package="com.uber.rib.workflow.test"/>
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -124,11 +124,6 @@ subprojects {
             // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
             project.android {
                 if (isAndroidLibrary) {
-                    libraryVariants.all {
-                        it.generateBuildConfigProvider.configure {
-                            it.enabled = false
-                        }
-                    }
                     variantFilter { variant ->
                         if (variant.buildType.name == 'debug') {
                             variant.setIgnore(true)
@@ -154,7 +149,7 @@ subprojects {
             def extraCompilerArgs = [
                     // See https://github.com/bazelbuild/bazel/issues/15144
                     "-Xjvm-default=enable",
-                    "-Xopt-in=kotlin.RequiresOptIn"
+                    "-opt-in=kotlin.RequiresOptIn",
             ]
             if (project.name in moduleFriends.keySet()) {
                 def friendName = moduleFriends[project.name]
@@ -175,20 +170,20 @@ subprojects {
                 tasks.withType(KotlinCompile).configureEach {
                     kotlinOptions {
                         freeCompilerArgs = extraCompilerArgs
-                        jvmTarget = "1.8"
+                        jvmTarget = "11"
                     }
                 }
             } else if(!isIntelliJPlugin) {
                 project.compileKotlin {
                     kotlinOptions {
                         freeCompilerArgs = extraCompilerArgs
-                        jvmTarget = "1.8"
+                        jvmTarget = "11"
                     }
                 }
                 project.compileTestKotlin {
                     kotlinOptions {
                         freeCompilerArgs = extraCompilerArgs
-                        jvmTarget = "1.8"
+                        jvmTarget = "11"
                     }
                 }
             }
--- a/android/demos/compose/build.gradle
+++ b/android/demos/compose/build.gradle
@@ -3,12 +3,12 @@ apply plugin: 'org.jetbrains.kotlin.android'
 apply plugin: "org.jetbrains.kotlin.kapt"
 
 android {
-    compileSdkVersion deps.build.compileSdkVersion
-    buildToolsVersion deps.build.buildToolsVersion
+    namespace "com.uber.rib.compose"
+    compileSdk deps.build.compileSdkVersion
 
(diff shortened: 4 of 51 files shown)


android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -157,10 +158,13 @@ abstract class RibActivity :
     callbacksFlow.tryEmit(createTrimMemoryEvent(level))
   }
 
+  @RequiresApi(api = Build.VERSION_CODES.O)
+  @CallSuper
   override fun onPictureInPictureModeChanged(
     isInPictureInPictureMode: Boolean,
     newConfig: Configuration,
   ) {
+    super.onPictureInPictureModeChanged(isInPictureInPictureMode, newConfig)
```
psteiger, line 167: @jbarr21 I think that's a new requirement because of the `androidx.activity` bump (I believe @CallSuper was added); from `super` source code I don't think there's any issues in calling super. 

Tagging you just in case you know something that I don't.

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("invisible_reference", "invisible_member")
```
tyvsmith, line 16: This is gross. Let's capture the reason by documenting youtrack issue. Also, did you see anything about a longer term direction on how to think about this? Are they not really supporting friendpaths anymore?

android/libraries/rib-workflow-test/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,2 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="com.uber.rib.workflow.test"/>
+<manifest />
```
tyvsmith, line 2: Are safe to kill otherwise empty manifests?

jbarr21 (approved):

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
psteiger: Agreed this is horrible. After much digging I think friend paths was never properly publicly supported/documented. I found it really hard to find any docs on this, even for the barebones `kotlinc` compiler. I basically dig into a lot of YouTrack issues to see the state of art on this and most issues are multiple years old. Not to mention Kotlin Gradle plugin. @jbarr21 might have some context on this as the original implementor of the friends Gradle config we use.

The alternatives I see are:
1. Don't access internal from other modules. But making stuff like `LazyCoroutineScope` public is a no-no.
2. (Untested, might be missing something) rethink module structure; create a third module for containing common components as `public` but don't expose it: import it as `implementation` in dependent modules. I'm just unsure if it's possible to do without generating public artifacts for this "third module", which we don't want. I think no.

So basically I don't like either strategy.

android/libraries/rib-workflow-test/src/main/AndroidManifest.xml
psteiger: Let me check and get back on this

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineScopes.kt
psteiger: Maybe there's some nice syntax for configuring friend paths in Gradle / Kotlin Gradle plugin as we do on BUCK and I'm just missing it :)

android/libraries/rib-workflow-test/src/main/AndroidManifest.xml
psteiger: I think this is recently allowed. [1](https://issuetracker.google.com/issues/242580918), [2](https://issuetracker.google.com/issues/232803637), [3](https://issuetracker.google.com/issues/232420210)

Will patch

idanakav (approved):
I'll take a look to see why the test is failing now, in the meantime, let's proceed, it shouldn't block this change.

tyvsmith (approved):

PR #572, opened 2023-05-02
Open `lifecycleFlow`, thus enabling it for mocking

psteiger:
Also, finalizes all `LifecycleScopeProvider` methods in order to impede mocking or overriding. Those methods will have `lifecycleFlow` as the source of truth (either real or mocked).

### DIFF
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("invisible_reference", "invisible_member")
+
 package com.uber.rib.core
 
 import android.R
@@ -23,8 +25,8 @@ import android.view.ViewGroup
 import androidx.annotation.CallSuper
 import com.uber.autodispose.lifecycle.CorrespondingEventsFunction
 import com.uber.autodispose.lifecycle.LifecycleEndedException
+import com.uber.autodispose.lifecycle.LifecycleNotStartedException
 import com.uber.autodispose.lifecycle.LifecycleScopeProvider
-import com.uber.autodispose.lifecycle.LifecycleScopes
 import com.uber.rib.core.lifecycle.ActivityCallbackEvent
 import com.uber.rib.core.lifecycle.ActivityCallbackEvent.Companion.create
 import com.uber.rib.core.lifecycle.ActivityCallbackEvent.Companion.createNewIntent
@@ -40,6 +42,7 @@ import io.reactivex.CompletableSource
 import io.reactivex.Observable
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.rx2.asObservable
 
 /** Base implementation for all VIP [android.app.Activity]s. */
@@ -49,37 +52,36 @@ abstract class RibActivity :
   LifecycleScopeProvider<ActivityLifecycleEvent>,
   RxActivityEvents {
   private var router: ViewRouter<*, *>? = null
-  private val lifecycleFlow =
+  private val _lifecycleFlow =
     MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  private val callbacksFlow =
+  open val lifecycleFlow: SharedFlow<ActivityLifecycleEvent>
+    get() = _lifecycleFlow
+  private val _callbacksFlow =
     MutableSharedFlow<ActivityCallbackEvent>(0, 1, BufferOverflow.DROP_OLDEST)
-  private val lifecycleObservable = lifecycleFlow.asObservable()
-  private val callbacksObservable = callbacksFlow.asObservable()
+  open val callbacksFlow: SharedFlow<ActivityCallbackEvent>
+    get() = _callbacksFlow
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun lifecycle(): Observable<ActivityLifecycleEvent> = lifecycleObservable
+  final override fun lifecycle(): Observable<ActivityLifecycleEvent> = lifecycleFlow.asObservable()
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun callbacks(): Observable<ActivityCallbackEvent> = callbacksObservable
+  override fun callbacks(): Observable<ActivityCallbackEvent> = callbacksFlow.asObservable()
 
-  override fun correspondingEvents(): CorrespondingEventsFunction<ActivityLifecycleEvent> {
-    return ACTIVITY_LIFECYCLE
-  }
+  final override fun correspondingEvents(): CorrespondingEventsFunction<ActivityLifecycleEvent> =
+    ACTIVITY_LIFECYCLE
 
-  override fun peekLifecycle(): ActivityLifecycleEvent? {
-    return lifecycleFlow.replayCache.lastOrNull()
-  }
+  final override fun peekLifecycle(): ActivityLifecycleEvent? =
+    lifecycleFlow.replayCache.lastOrNull()
 
-  override fun requestScope(): CompletableSource {
-    return LifecycleScopes.resolveScopeFromLifecycle(this)
-  }
+  final override fun requestScope(): CompletableSource =
+    lifecycleFlow.asScopeCompletable(lifecycleRange)
 
   @Initializer
   @CallSuper
   override fun onCreate(savedInstanceState: android.os.Bundle?) {
     super.onCreate(savedInstanceState)
     val rootViewGroup = findViewById<ViewGroup>(R.id.content)
-    lifecycleFlow.tryEmit(createOnCreateEvent(savedInstanceState))
+    _lifecycleFlow.tryEmit(createOnCreateEvent(savedInstanceState))
     val wrappedBundle: Bundle? =
       if (savedInstanceState != null) Bundle(savedInstanceState) else null
     router = createRouter(rootViewGroup)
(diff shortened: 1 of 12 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Presenter.kt
```diff
@@ -69,9 +64,9 @@ public abstract class Presenter : ScopeProvider {
   @UIEffect @CallSuper protected open fun willUnload() {}
 
   /** @return an observable of this controller's lifecycle events. */
-  public open fun lifecycle(): Observable<PresenterEvent> = lifecycleObservable
+  public fun lifecycle(): Observable<PresenterEvent> = lifecycle.asObservable()
```
tyvsmith, line 67: final?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -30,44 +29,42 @@ import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.rx2.asObservable
+import org.jetbrains.annotations.TestOnly
```
tyvsmith, line 32: Unneeded now.

tyvsmith:
If we're exposing Flow APIs for these classes to get tests working, we should probably do it for the remainder of ones that expose Observables directly and use backing Flow fields, even if we don't have tests failing.

RibActivity, ViewProvider, RouterNavigatorEvents, RibEvents.

Then following that, I think we can probably add .cortouineScope directly to these lifecycle classes as opposed to extension function on ScopeProvider

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Presenter.kt
psteiger: Let's do it

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -37,52 +36,49 @@ import kotlinx.coroutines.rx2.asObservable
...
+public abstract class Interactor<P : Any, R : Router<*>>() : InteractorType {
   @Inject public lateinit var injectedPresenter: P
   internal var actualPresenter: P? = null
   private val _lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-
-  // We expose lifecycleFlow internally to avoid some unnecessary Rx-Coroutine interop. Use it
-  // instead of `lifecycle()`.
-  @get:JvmSynthetic
-  internal val lifecycleFlow: SharedFlow<InteractorEvent>
+  public open val lifecycleFlow: SharedFlow<InteractorEvent>
```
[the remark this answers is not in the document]
FranAguilera: It's strange it's working on my Android studio version (Android Studio Flamingo | 2022.2.1)

image

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Presenter.kt
```diff
@@ -69,9 +64,10 @@ public abstract class Presenter : ScopeProvider {
   @UIEffect @CallSuper protected open fun willUnload() {}
 
   /** @return an observable of this controller's lifecycle events. */
-  public open fun lifecycle(): Observable<PresenterEvent> = lifecycleObservable
```
FranAguilera, line 72: Nit: Same as above

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -49,37 +52,36 @@ abstract class RibActivity :
   LifecycleScopeProvider<ActivityLifecycleEvent>,
   RxActivityEvents {
   private var router: ViewRouter<*, *>? = null
-  private val lifecycleFlow =
+  private val _lifecycleFlow =
     MutableSharedFlow<ActivityLifecycleEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  private val callbacksFlow =
+  open val lifecycleFlow: SharedFlow<ActivityLifecycleEvent>
```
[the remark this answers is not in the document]
tyvsmith: Probably not just for testing. Open for it, yes. But visible as part of the first class Flow APIs

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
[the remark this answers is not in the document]
psteiger: Did it work on IDEA ? 
Nevertheless, it seems the API is now meant to be public as per @tyvsmith , it's just "Open for tests" and AFAIK there's no annotation for such case. Please reopen if you disagree or have any other suggestion

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -49,37 +52,36 @@ abstract class RibActivity :
...
+  private val _callbacksFlow =
     MutableSharedFlow<ActivityCallbackEvent>(0, 1, BufferOverflow.DROP_OLDEST)
-  private val lifecycleObservable = lifecycleFlow.asObservable()
-  private val callbacksObservable = callbacksFlow.asObservable()
+  open val callbacksFlow: SharedFlow<ActivityCallbackEvent>
+    get() = _callbacksFlow
 
   /** @return an observable of this activity's lifecycle events. */
-  override fun lifecycle(): Observable<ActivityLifecycleEvent> = lifecycleObservable
+  final override fun lifecycle(): Observable<ActivityLifecycleEvent> = lifecycleFlow.asObservable()
```
tyvsmith, line 65: How about a lateinit backing var to avoid doing this on every call.  Or we can make a new extension function like asCachedObservable() where we hold onto it?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -49,37 +52,36 @@ abstract class RibActivity :
...
-    return lifecycleFlow.replayCache.lastOrNull()
-  }
+  final override fun peekLifecycle(): ActivityLifecycleEvent? =
+    lifecycleFlow.replayCache.lastOrNull()
 
-  override fun requestScope(): CompletableSource {
-    return LifecycleScopes.resolveScopeFromLifecycle(this)
-  }
+  final override fun requestScope(): CompletableSource =
+    lifecycleFlow.asScopeCompletable(lifecycleRange)
```
tyvsmith, line 77: Why this one?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
```diff
@@ -35,6 +35,8 @@ private constructor(
     DESTROY,
   }
 
+  override fun compareTo(other: ActivityLifecycleEvent): Int = type.compareTo(other.type)
```
tyvsmith, line 38: Why was this one needed?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -241,3 +243,10 @@ abstract class RibActivity :
       }
   }
 }
+
+private val <T : Comparable<T>> LifecycleScopeProvider<T>.lifecycleRange: ClosedRange<T>
+  get() {
```
tyvsmith, line 248: Is the idea to just catch the error earlier than when it's launcehd in unstarted ones? Do we need to handle other complexity of the original Autodispose resolveScopes method?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
psteiger: Hmm... Yeah, I think we might have an opportunity of "lazy" here

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
psteiger: Plus, -1 dependency on `autodispose`.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
psteiger: I believe it's answered above.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
psteiger: Yes! We must throw at call site to keep the behavior. I don't thought thoroughly whether we *should*, mostly worried about retro-compatibility, but this one may have good reasons to throw at call site instead of in `Completable`.

We also throw early if pass in a range of active ranging from, e.g. (`STARTED`..`STOPPED`) but `SharedFlow` is still in `CREATED` (note on current implementation it should never happen because we start the range with `peekLifecycle()`):
```kotlin
internal fun <T : Comparable<T>> SharedFlow<T>.asScopeCompletable(
  range: ClosedRange<T>,
  context: CoroutineContext = EmptyCoroutineContext,
): CompletableSource {
  ensureAlive(range)
  return rxCompletable(RibDispatchers.Unconfined + context) { first { it == range.endInclusive } }
}

private fun <T : Comparable<T>> SharedFlow<T>.ensureAlive(range: ClosedRange<T>) {
  val lastEmitted = replayCache.lastOrNull()
  when {
    lastEmitted == null || lastEmitted < range.start -> throw LifecycleNotStartedException()
    lastEmitted >= range.endInclusive -> throw LifecycleEndedException()
  }
}
```

psteiger: > Do we need to handle other complexity of the original Autodispose resolveScopes method?

I don't think so. I think we got this covered. As I mentioned, it's a hopefully less complex impl that achieves the same behavior (and does not depend on Flow->Observable interop).

PR #573, opened 2023-05-02
[WorkerBinder] Guard against potential Worker.coroutineContext being null while using Mockito

FranAguilera:
**Description**:

Tests that internally bind workers on Interactor's/Presenter and provide its Worker dependency via a mock (in Mockito) will start failing due to Mockito changing the default return type of `Worker.coroutineContext` from `EmptyCoroutineContext` to a non-possible null type otherwise.

This PR just guards against this possibility for testing and extract a function to easily get this context at binding logic in a safer manner

**Related issue(s)**:

https://github.com/uber/RIBs/issues/574

Expanded test on InteractorAndRouterTest. Will add few more test cases on new Workers demo module as well

### DIFF
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -34,6 +34,13 @@ import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.takeWhile
 import kotlinx.coroutines.launch
 
+/**
+ * Resulting CoroutineContext defined at [Worker] that guards against potential nullable cases on
+ * tests that have a mocked Worker via Mockito
+ */
+private val Worker.bindingCoroutineContext: CoroutineContext
+  get() = this.coroutineContext ?: EmptyCoroutineContext
+
 /** Helper class to bind to an interactor's lifecycle to translate it to a [Worker] lifecycle. */
 public object WorkerBinder {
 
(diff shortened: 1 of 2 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Worker.kt
```diff
@@ -56,4 +56,17 @@ public interface Worker {
...
+
+  public companion object {
+
+    /**
+     * Returns the CoroutineContext defined at [Worker].
+     *
+     * In testing environments that rely on mockito, a Mocked [Worker] will return a null
+     * coroutineContext.
+     */
+    public fun Worker.getBindingContext(): CoroutineContext {
```
psteiger, line 68: I prefer it private, and inside `WorkerBinder` file at top level. Why do we need to expose this ext fn?

FranAguilera: oh that's right, we discussed internally. Let me update in a bit

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -37,6 +38,9 @@ import kotlinx.coroutines.launch
 /** Helper class to bind to an interactor's lifecycle to translate it to a [Worker] lifecycle. */
 public object WorkerBinder {
 
+  internal val Worker.bindingCoroutineContext: CoroutineContext
```
psteiger, line 41: Can't it be `private` ? I don't see it being used in any other file

psteiger: Oh, guess it's used in the listener

FranAguilera: it can't, given the main bind method is defined outside WorkerBinder

psteiger, line 41: I'd prefer to have it as top level in the file (outside WorkerBinder object). I suspect then it can be `private`. If it can't be private even if top-level, consider it as nit and I'll approve as is

FranAguilera: Oh I see, yes definitely can move it up

psteiger (approved):

PR #575, opened 2023-05-03
Add  Rib Worker demo app

FranAguilera:
**Description**:

Add basic demo app with different WorkerBinder options and default monitoring

image

image


android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/RootActivity.kt
```diff
@@ -0,0 +1,45 @@
...
+import motif.Creatable
+import motif.Expose
+import motif.NoDependencies
+import motif.ScopeFactory
+
+class RootActivity : RibActivity() {
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    WorkerBinder.initializeMonitoring { println(it) }
```
FranAguilera, line 32: WorkerBinder monitoring initialization here

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/main/ribworkerselection/RibWorkerSelectionInteractor.kt
```diff
@@ -0,0 +1,100 @@
...
+
+  override fun didBecomeActive(savedInstanceState: Bundle?) {
+    super.didBecomeActive(savedInstanceState)
+    eventStream
+      .observe()
+      .onEach {
+        when (it) {
+          RibWorkerBindTypeClickType.SINGLE_WORKER_BIND_CALLER_THREAD -> {
+            updateViewModel(uiWorker::class.simpleName)
+            WorkerBinder.bind(this, uiWorker)
```
FranAguilera, line 56: Different binding types

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/main/workers/ExpensiveFunctionSimulator.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.workers.root.main.workers
+
+object ExpensiveFunctionSimulator {
+
+  fun runExpensiveFunction() {
+    val duration = (100..250).random().toLong()
+    Thread.sleep(duration)
```
tyvsmith, line 22: Use delay(), the kotlin way, not coupled to jvm.

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/main/workers/UiWorker.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.workers.root.main.workers
+
+import com.uber.rib.core.RibDispatchers
+import com.uber.rib.core.Worker
+import kotlin.coroutines.CoroutineContext
+
+class UiWorker : Worker {
```
tyvsmith, line 22: Good overall, but it still leaves the concept of a worker as pretty ambiguous, with most of these being empty or just a sleep. We should try to flesh these out a bit more to teach the concept.

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/RootActivity.kt
FranAguilera: Good point, will defined a more real case sample of monitoring (e.g. warn on debug builds of long UI durations) and initialized at RootInteractor.

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/main/workers/UiWorker.kt
FranAguilera: Makes sense, will add more details for this type of workers. E.g. Io Worker for some file manipulation, etc

tyvsmith (approved):

PR #581, opened 2023-05-11
Update LeakCanary to 2.10

pyricau:
Also added the now required "description" parameter when watching instances.

# Before

image (56)

# After

image (55)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
```diff
@@ -141,7 +141,7 @@ protected constructor(
   public open fun detachChild(childRouter: Router<*>) {
     val isChildRemoved = children.remove(childRouter)
     val interactor = childRouter.interactor
-    ribRefWatcher.watchDeletedObject(interactor)
+    ribRefWatcher.watchDeletedObject(interactor, "detached child router ${childRouter.javaClass.name}")
```
pyricau, line 144: I made a description. You might want something different. But it's nice to be able to provide context on what's causing this to be defined as something to be GCed.

android/demos/memory-leaks/src/main/java/com/uber/rib/SampleApplication.java
```diff
@@ -31,24 +29,17 @@ public class SampleApplication extends Application implements HasActivityDelegat
   public void onCreate() {
     activityDelegate = new SampleActivityDelegate();
     super.onCreate();
-    if (!LeakCanary.isInAnalyzerProcess(this)) {
```
tyvsmith, line 34: Was the separate process deprecated?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibRefWatcher.kt
```diff
@@ -39,12 +39,15 @@ public open class RibRefWatcher {
    *
    * @param objectToWatch the object to watch.
    */
-  public open fun watchDeletedObject(objectToWatch: Any?) {
+  public open fun watchDeletedObject(
+    objectToWatch: Any?,
+    description: String
```
tyvsmith, line 44: Should be an optional param to avoid being a breaking API change for consumers

android/demos/memory-leaks/src/main/java/com/uber/rib/SampleApplication.java
pyricau: It's optional now. Most apps shouldn't need it as LeakCanary uses a lot less memory, and this complicates the setup.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibRefWatcher.kt
pyricau: Done by adding a deprecated overload without the param (that way it signifies clearly that the call should be updated)

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibRefWatcher.kt
```diff
@@ -34,17 +34,30 @@ public open class RibRefWatcher {
...
 
+  @Deprecated(
+    "Add the description parameter",
+    replaceWith = ReplaceWith("watchDeletedObject(objectToWatch, description)"),
+  )
+  public open fun watchDeletedObject(
+    objectToWatch: Any?,
+  ) {
+    watchDeletedObject(objectToWatch, "missing description")
+  }
```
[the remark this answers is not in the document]
tyvsmith: Looks good. Thanks

PR #583, opened 2023-05-14
Make use of `jvmToolchain` for building the project.

### LINKED ISSUE #584: Use JVM toolchain for building the project
When building the RIBs Android project on my personal machine, which has `JAVA_HOME` set to a Amazon Corretto 17 JDK, I encounter the following error:
```
Execution failed for task ':libraries:rib-base:kaptGenerateStubsKotlin'.
> 'compileJava' task (current target is 11) and 'kaptGenerateStubsKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

We should be using [JVM Toolchain](https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support) for avoiding such issues, making the build process more predictable and repeatable across different local configurations.

psteiger:
This allows for a more repeatable, predictable build between different local configurations.

On Kotlin modules, this is achieved by adding the following to project-level `build.gradle`:
```
kotlin {
  jvmToolchain(11)
}
```

On Java modules,

```
java {
  toolchain {
    languageVersion.set(JavaLanguageVersion.of(11))
  }
}
```

For non-Android Kotlin, when using the JVM Toolchain, the following is not needed anymore:
```
sourceCompatibility = deps.build.javaVersion.toString()
targetCompatibility = deps.build.javaVersion.toString()
```

For Android Kotlin modules, because of [a bug present in AGP 7.4.x up to 8.1.0-alpha08](https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support), the following is still needed:
```
android {
    compileOptions {
        sourceCompatibility = <MAJOR_JDK_VERSION>
        targetCompatibility = <MAJOR_JDK_VERSION>
    }
}
```

Because AGP is bundled with the Android IntelliJ IDEA plugin, we cannot upgrade it (maximum supported version is 7.4 on 2023.1.1). We should open a new issue for removing above configuration block when a newer stable IntelliJ IDEA version ships with support for AGP 8.1.0.alpha09 or higher

Fixes #584

### DIFF
--- a/android/libraries/rib-android-compose/build.gradle
+++ b/android/libraries/rib-android-compose/build.gradle
@@ -18,6 +18,11 @@ apply plugin: 'com.android.library'
 apply plugin: 'org.jetbrains.kotlin.android'
 apply plugin: "com.vanniktech.maven.publish"
 
+kotlin {
+    explicitApi()
+    jvmToolchain(11)
+}
+
 android {
     namespace "com.uber.rib.android.compose"
     compileSdk deps.build.compileSdkVersion
@@ -52,8 +57,4 @@ dependencies {
     testImplementation deps.external.roboelectricBase
     testImplementation deps.test.mockitoKotlin
     testImplementation project(":libraries:rib-test")
-}
-
-kotlin {
-    explicitApi()
 }
--- a/android/demos/compose/build.gradle
+++ b/android/demos/compose/build.gradle
@@ -2,6 +2,10 @@ apply plugin: 'com.android.application'
 apply plugin: 'org.jetbrains.kotlin.android'
 apply plugin: "org.jetbrains.kotlin.kapt"
 
+kotlin {
+    jvmToolchain(11)
+}
+
 android {
     namespace "com.uber.rib.compose"
     compileSdk deps.build.compileSdkVersion
--- a/android/demos/flipper/build.gradle
+++ b/android/demos/flipper/build.gradle
@@ -22,6 +22,12 @@ buildscript {
 
 apply plugin: 'com.android.application'
 
+java {
+    toolchain {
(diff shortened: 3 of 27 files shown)


android/libraries/rib-android-compose/build.gradle
```diff
@@ -18,6 +18,11 @@ apply plugin: 'com.android.library'
 apply plugin: 'org.jetbrains.kotlin.android'
 apply plugin: "com.vanniktech.maven.publish"
 
+kotlin {
```
tyvsmith, line 21: We should probably apply this by default to all build targets (with the couple exceptions until we convert/fix them), so this isn't managed in ~20 spots.

psteiger: Agreed with the general statement of centralizing Kotlin module configurations. I can try to move those two to the project-level build.gradle.

With that said, the reason I did not do it on our project-level `build.gradle` is IMO we do config centralization poorly, in a non-Gradle-idiomatic way. We should probably be using a Gradle conventions plugin instead of those hard-to-read (to me) configurations at project-level, either built from `buildSrc` or `includedBuilds`, then we could just apply the plugin like
```
plugins {
 id("ribs-kotlin-convention-plugin")
}
```

This [seems to be](https://docs.gradle.org/current/samples/sample_convention_plugins.html) the Gradle-way for applying common configurations between different modules.

Perhaps we could open an issue to improve on that.

tyvsmith: Agreed. Let's punt this, and create an issue for the plugin, as well as moving the remaining java modules over and enabling explicitApi on the outlier.

tyvsmith (approved):

PR #586, opened 2023-05-16
Improve KDoc on `ActivityLifecycleEvent` by explaining ordering semantics.

psteiger:
Also, a couple of minor fixes:
- A small warning on `RibActivity` against importing `android.R`.
- Replaces deprecated `String.toLowerCase()` and `String.capitalize()` with the equivalent `String.lowercase()` and `String.replaceFirstChar(Char::titlecase)`.

### DIFF
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
@@ -17,7 +17,34 @@ package com.uber.rib.core.lifecycle
 
 import android.os.Bundle
 
-/** Lifecycle events that can be emitted by Activities. */
+/**
+ * Lifecycle events that can be emitted by Activities.
+ *
+ * ### Ordering semantics
+ *
+ * This class implements [Comparable], but it does **not** override `equals`, so even though the
+ * following holds:
+ * ```
+ * val a = createOnCreateEvent(null)
+ * val b = createOnCreateEvent(Bundle())
+ * assertThat(a <= b).isTrue()
+ * assertThat(b <= a).isTrue()
+ * ```
+ *
+ * The equality does not hold:
+ * ```
+ * assertThat(a == b).isFalse()
+ * ```
+ *
+ * This happens because events of type [CREATE][ActivityLifecycleEvent.Type.CREATE] hold a [Bundle],
+ * and even though any two `CREATE` events are equal in ordering ([compareTo]), they are never equal
+ * in [equals] comparison.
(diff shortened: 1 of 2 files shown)


android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
```diff
@@ -17,7 +17,34 @@ package com.uber.rib.core.lifecycle
...
+ * ```
+ * val a = createOnCreateEvent(null)
+ * val b = createOnCreateEvent(Bundle())
+ * assertThat(a <= b).isTrue()
+ * assertThat(b <= a).isTrue()
+ * ```
+ *
+ * The equality does not hold:
+ * ```
+ * assertThat(a == b).isFalse()
```
tyvsmith, line 36: Should they be?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
```diff
@@ -79,7 +109,7 @@ private constructor(
         Type.DESTROY -> DESTROY_EVENT
         else ->
           throw IllegalArgumentException(
-            "Use the createOn${type.name.toLowerCase().capitalize()}Event() method for this type!",
+            "Use the createOn${type.name.lowercase().replaceFirstChar(Char::titlecase)}Event() method for this type!",
```
tyvsmith, line 112: Kind of surprised they replaced capitalize with this, it's much more verbose and less obvious for folks.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
psteiger: IMO, no, because Android bundles also do not implement equal

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/lifecycle/ActivityLifecycleEvent.kt
psteiger: Actually the drop in replacement is even more verbose (suggested by ReplaceWith).

From googling It, it seems the original deprecated capitalize function had issues with
1. Not taking a locale 
2. Surprising behavior on characters that are one char in Unicode, but actually two chars in the natural language.

Both of which don't matter at this particular call site of ours

tyvsmith (approved):

PR #592, opened 2023-05-25
Use a direct dispatcher in `Flow` -> `Rx` conversions.

psteiger:
A direct dispatcher mimics the Rx always synchronous-by-default behavior (unless `subscribeOn`/`observeOn` is used), which is different to the `Dispatchers.Unconfined` behavior (default used by `Flow.asObservable`) which may form event loops for nested coroutines (asynchronous).


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/DirectDispatcher.kt
```diff
@@ -0,0 +1,38 @@
...
+ * coroutines.
+ *
+ * For more context, see the following issues on `kotlinx.coroutines` GitHub repository:
+ * <!-- spotless:off -->
+ * 1. [Immediate dispatchers can cause spooky action at a distance](https://github.com/Kotlin/kotlinx.coroutines/issues/3760)
+ * 2. [Chaining rxSingle calls that use Unconfined dispatcher and blockingGet results in deadlock #3458](https://github.com/Kotlin/kotlinx.coroutines/issues/3458)
+ * 3. [Coroutines/Flow vs add/remove listener (synchronous execution) #3506](https://github.com/Kotlin/kotlinx.coroutines/issues/3506)
+ * <!-- spotless:on -->
+ */
+internal object DirectDispatcher : CoroutineDispatcher() {
```
FranAguilera, line 34: thanks for adding! wondering if this should be moved into another module (e.g. libraries/rib-coroutines)

FranAguilera:
What do you think if we update WorkerBinder default param value with Unconfined with this new DirectDispatcher? https://github.com/uber/RIBs/blob/main/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt#L78

psteiger:
@FranAguilera that's a good question 

I think we would be wise to do some benchmarks to measure performance impact of this change and also experiment internally on the safety (stackoverflow risk etc)

FranAguilera:
> @FranAguilera that's a good question
> 
> I think we would be wise to do some benchmarks to measure performance impact of this change and also experiment internally on the safety (stackoverflow risk etc)

@psteiger Sorry missed this comment earlier. I totally agree, we can put internal doc for steps on measuring impact of this change within our internal repo

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/DirectDispatcher.kt
FranAguilera: @psteiger friendly ping here ^

psteiger: Sorry I missed this comment.

I think it's best to have it in "core", because I believe the tendency is for rib-coroutine to cease to exist once we migrate fully to coroutines.

@tyvsmith in case you wanna confirm or deny ^

FranAguilera: np! I see, thanks for clarifying

tyvsmith: Lets move it to ribs-coroutines until we're ready to move all of it to core. Otherwise we're in a worse fragmented state.

PR #594, opened 2023-06-07
Deprecate existing WorkerBinder.bind methods and provide recommended bindTo methods with RibDispatchers.Default

FranAguilera:
**Description**:

Since https://github.com/uber/RIBs/pull/553 we added support to specify on which CoroutineDispatcher are presidio Worker will be bound. RibDispatchers.Unconfined was added as a default parameter in order to keep existing behavior and not be a breaking change. 

Still, binding by default on the current caller thread within interactor means that by default we are binding workers on main UI thread leading to potential jank ANR etc just for binding those workers. 

This is specially problematic when the list of workers being bound keeps increasing (e.g. hypothetically if we have 100 workers and each bind takes at least 50 ms this will result on the UI thread being blocked for at least 5000ms)

Proposal is to deprecate existing methods and provide new methods that by default will bind on **Ribdispatchers.DEFAULT** (similarly to **RibCoroutineWorker**). Still if there is a need for a Worker to change the thread where it will be bound we can rely on **coroutineContext** property to be overriden as following (default value is **EmptyCoroutineContext**):

```
  class MyUiWorker:com.uber.rib.core.Worker{
    override val coroutineContext: CoroutineContext = RibDispatchers.Main
  }
```

Demo call:

`worker.bindTo(interactor)
`

NOTE: For calling on existing java files -> `WorkerBinderKt.bindTo(ribWorker, this));`

**Related issue(s)**:

https://github.com/uber/RIBs/issues/595

Verified via Unit tests

### DIFF
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -62,11 +62,12 @@ public object WorkerBinder {
   }
 
   /**
-   * Initializes [WorkerBinderMigrationProvider]
+   * Initializes [WorkerBinderMigrationProvider]. It will be used to allow changing the current
+   * default CoroutineDispatchers being bound from [RibDispatchers.Main] to [RibDispatchers.Default]
+   * at WorkerBinder.bind calls.
    *
-   * Provide a concrete implementation of [WorkerBinderMigrationProvider] that allow controlled
-   * changes (e.g. via remote configuration) for changing current default CoroutineDispatchers being
-   * bound from [RibDispatchers.Main] to [RibDispatchers.Default]
+   * This could be used within a remote configuration that will allow for a safe revert in case of
+   * issues detected (e.g. Specific Worker must be bound on main thread)
    *
    * IMPORTANT: If set, this should be called at the earliest scope possible to allow migrating
    * early workers
@@ -447,13 +448,18 @@ public fun interface WorkerBinderListener {
  * allow a safe migration (e.g. via remove config) and allowing to change the current
  * CoroutineDispatcher being used in WorkerBinder.bind calls from [CoroutineDispatchers.Unconfined]
  * to [RibDispatchers.Default]
+ *
+ * IMPORTANT NOTE: For calls to WorkerBinder.bind that receive a list of workers and one of the
+ * workers in the list requires to be bound on main thread, a concrete override can be applied on
+ * the Worker itself (e.g. by overriding default coroutineContext at Worker from
+ * [EmptyCoroutineContext] to [RibDispatchers.Main]
  */
 public fun interface WorkerBinderMigrationProvider {
 
   /**
-   * Returns true with true when the original default CoroutineDispatcher at WorkerBinder should be
-   * migrated to a background dispatcher (e.g. from [CoroutineDispatchers.Unconfined] to
-   * [RibDispatchers.Default])
+   * Returns true when the default CoroutineDispatcher ([CoroutineDispatchers.Unconfined]) at
+   * WorkerBinder should be migrated to a background dispatcher (e.g. from
+   * [CoroutineDispatchers.Unconfined] to [RibDispatchers.Default])
    */
   public fun shouldMigrateDispatcherAtWorkerBinder(): Boolean
 }
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -72,7 +72,7 @@ public object WorkerBinder {
    * early workers
    */
   @JvmStatic
-  public fun initializeDispatcherMigration(
+  public fun initializeWorkerBinderMigration(
     workerBinderMigrationProvider: WorkerBinderMigrationProvider,
   ) {
     this.workerBinderMigrationProvider = workerBinderMigrationProvider
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -61,6 +61,46 @@ public object WorkerBinder {
 
   /**
    * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
+   * interactor's lifecycle events
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.DEFAULT] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * Inject this class into your interactor and call this method on any
+   *
+   * @param interactor The interactor that provides the lifecycle.
+   * @param worker The class that wants to be informed when to start and stop doing work.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun bindToInteractor(
+    interactor: Interactor<*, *>,
+    worker: Worker,
+  ): WorkerUnbinder = bind(interactor, worker, RibDispatchers.Default)
+
+  /**
+   * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
+   * interactor's lifecycle events
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.DEFAULT] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * Inject this class into your interactor and call this method on any
+   *
+   * @param interactor The interactor that provides the lifecycle.
+   * @param workers A list of classes that want to be informed when to start and stop doing work.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun bindToInteractor(
+    interactor: Interactor<*, *>,
+    workers: List<Worker>,
+  ) {
+    bind(interactor, workers, RibDispatchers.Default)
+  }
+
+  /**
+   * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
    * interactor's lifecycle events. Inject this class into your interactor and call this method on
    * any
    *
@@ -70,6 +110,14 @@ public object WorkerBinder {
    *   is not overriden with a value different that [EmptyCoroutineContext]
    * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
    */
+  @Deprecated(
+    message =
+      """
+      This method will bind on current thread where WorkerBinder is called. This often means that
+      the binding will happen on the main Thread leading to potential Jank issues/ANRs
+    """,
+    replaceWith = ReplaceWith("bindToInteractor(interactor, workers)"),
+  )
   @JvmStatic
   @JvmOverloads
   public fun bind(
@@ -95,6 +143,14 @@ public object WorkerBinder {
    *   [Worker.coroutineContext] is not overriden with a value different than
    *   [EmptyCoroutineContext]
    */
+  @Deprecated(
+    message =
+      """
+      This method will bind on current thread where WorkerBinder is called. This often means that
+      the binding will happen on the main Thread leading to potential Jank issues/ANRs
+    """,
+    replaceWith = ReplaceWith("bindToInteractor(interactor, workers)"),
+  )
   @JvmStatic
   @JvmOverloads
   public fun bind(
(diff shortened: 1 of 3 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -59,6 +59,46 @@ public object WorkerBinder {
     this.workerBinderListenerWeakRef = WeakReference<WorkerBinderListener>(workerBinderListener)
   }
 
+  /**
+   * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
+   * interactor's lifecycle events
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.DEFAULT] unless the Worker explicitly
```
psteiger, line 66: Non-caps `RibDispatchers.Default`

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -59,6 +59,46 @@ public object WorkerBinder {
...
+  ): WorkerUnbinder = bind(interactor, worker, RibDispatchers.Default)
+
+  /**
+   * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
+   * interactor's lifecycle events
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.DEFAULT] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * Inject this class into your interactor and call this method on any
```
psteiger, line 88: On any... ?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -59,6 +59,46 @@ public object WorkerBinder {
...
   }
 
+  /**
+   * Bind a worker (ie. a manager or any other class that needs an interactor's lifecycle) to an
+   * interactor's lifecycle events
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.DEFAULT] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * Inject this class into your interactor and call this method on any
```
psteiger, line 69: On any... ?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -59,6 +59,46 @@ public object WorkerBinder {
...
+   *
+   * @param interactor The interactor that provides the lifecycle.
+   * @param worker The class that wants to be informed when to start and stop doing work.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun bindToInteractor(
+    interactor: Interactor<*, *>,
+    worker: Worker,
```
psteiger, line 78: Do you feel like we should provide an optional `dispatcher` parameter like we had before, but defaulting to `RibDispatchers.Default` instead?

```
  @JvmStatic @JvmOverloads
  public fun bindToInteractor(
    interactor: Interactor<*, *>,
    worker: Worker,
    dispatcher: CoroutineDispatcher = RibDispatchers.Default,
  ): WorkerUnbinder = bind(interactor, worker, dispatcher)
```

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -107,6 +163,42 @@ public object WorkerBinder {
...
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * @param presenter The presenter that provides the lifecycle.
+   * @param worker The class that wants to be informed when to start and stop doing work.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun bindToPresenter(
+    presenter: Presenter,
+    worker: Worker,
```
psteiger, line 180: Default dispatcher parameter?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -107,6 +163,42 @@ public object WorkerBinder {
...
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * @param presenter The presenter that provides the lifecycle.
+   * @param workers A list of classes that want to be informed when to start and stop doing work.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun bindToPresenter(
+    presenter: Presenter,
+    workers: List<Worker>,
```
psteiger, line 197: Default dispatcher parameter?

psteiger (changes requested):

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: good catch, will update

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: uops, yeah I copy pasted from the previous javadoc but that was clearly off. Will update in a bit

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: I guess we could keep it but at the same time I'm not too sold on this given the worker can specify its own dispatcher to override the one passed at WorkerBinder. 

If there is a need to specify another dispatcher probably is best to define within Worker level for clarification and to prevent unwanted changes outside the worker definition

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: discussed above

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
psteiger: I'm fine with that

psteiger (approved):

FranAguilera:
Sorry about the back and forth @psteiger, decided to take advantage of this and provide a migration API for all current WorkerBinder.bind calls. Will share internal doc offline too

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -243,6 +389,19 @@ public object WorkerBinder {
       }
     }
   }
+
+  private fun getDispatcherFromMigration(
```
tyvsmith, line 393: The set of APIs around the migration feels like we're leaking to much internal business logic into the public APis of Ribs. We should generalize this more. Perhaps just one override where we can set a static overriding dispatcher.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -95,21 +168,74 @@ public object WorkerBinder {
...
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.Default] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * @param presenter The presenter that provides the lifecycle.
+   * @param workers A list of classes that want to be informed when to start and stop doing work.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun bindToPresenter(
```
tyvsmith, line 227: This looks like new functionality added that isn't needed for these changes (unless I missed it's usage) where we support binding to a list.

1. Should this be a separate PR if we want a new puiblic API.
2. Rather in this diff or on the next, probably rename to bindToPresenters (plural)

FranAguilera: We've been providing these APIs (_**WorkerBinder.bind(interactor, listOfWorkers)/ WorkerBinder.bind(presenter, listOfWorkers**_) for a while. 

Part of the goal of this PR is to deprecate existing WorkerBinder.bind methods given that the methods with 2 parameter call takes a default RibDispatchers.Unconfined which often times results in binding on main thread. 

Was thinking that adding new methods (bindToInteractor, bindToPresenter) that are opinionated on using to RibDispatchers.Default to prevent further additions of WorkerBinder.bind on main thread (we can easily rely on static analysis to force everyone to use the new API when new WorkerBinder additions). 

Also another important caveat, if there is still a need for a legacy worker to be bound on the main thread this can be achieved by overriding Worker's coroutineContext property (this will ignore the default Dispatchers specified on WorkerBinder call)

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: I agree but also I'm a bit concerned that providing an easy static dispatcher override could lead to unwanted dispatchers change at any point by any consumers.

Let me think about this a bit more

FranAguilera: @tyvsmith Decided to keep this migration in internal monorepo. This PR will just focus on deprecating existing API and add equivalent new methods for the default binding happening off the main thread

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -59,17 +59,64 @@ public object WorkerBinder {
...
+   * interactor's lifecycle events
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.Default] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * @param interactor The interactor that provides the lifecycle.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun Worker.bindTo(
```
psteiger, line 73: +1 to extension function. 

But since we're providing those new APIs as public extension function, I'd just remove them from inside the `WorkerBinder` object.

Consider it a nit.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -108,8 +163,48 @@ public object WorkerBinder {
...
+   * presenter's lifecycle events.
+   *
+   * IMPORTANT: Binding will happen on [RibDispatchers.Default] unless the Worker explicitly
+   * overrides [coroutineContext] to other than [EmptyCoroutineContext]
+   *
+   * @param presenter The presenter that provides the lifecycle.
+   * @return [WorkerUnbinder] to unbind [Worker]'s lifecycle.
+   */
+  @JvmStatic
+  public fun Worker.bindTo(
```
psteiger, line 176: Same

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
FranAguilera: good catch! will be moving this outside WorkerBinder object in a sec

FranAguilera: @psteiger Decided to move all existing extensions to its own file so we can call the new extension functions on legacy java code

FranAguilera: Please ignore this. Reverted back to current WorkerBinder

FranAguilera:
> I don't have any conceptual objections. Exposing new API as public top-level function instead of nested inside `WorkerBinder` object is nice. And deprecating the previous API instead of changing it keeps retro-compat.
> 
> With that said, maybe we should consider deprecating `Worker` in favor of `RibCoroutineWorker`. I'm not sure it's worth to keep evolving Worker API. But if it's being done as a smoother transition pathway, I'm ok with it.

I completely agree with eventually fully deprecating this old worker. The only missing piece for RibCoroutineWorker is monitoring option for binding duration (but given that by default is bound on RibDispatchers.Default. there is no much of a concern). Also we can validate a few of this RibCoroutineWorkers in prod and if no issues found we can completely mark worker as deprecated

FranAguilera:
> > I don't have any conceptual objections. Exposing new API as public top-level function instead of nested inside `WorkerBinder` object is nice. And deprecating the previous API instead of changing it keeps retro-compat.
> > With that said, maybe we should consider deprecating `Worker` in favor of `RibCoroutineWorker`. I'm not sure it's worth to keep evolving Worker API. But if it's being done as a smoother transition pathway, I'm ok with it.
> 
> I completely agree with eventually fully deprecating this old worker. The only missing piece for RibCoroutineWorker is monitoring option for binding duration (but given that by default is bound on RibDispatchers.Default. there is no much of a concern). Also we can validate a few of this RibCoroutineWorkers in prod and if no issues found we can completely mark worker as deprecated

Actually, after thinking about a bit more I don't think we should keep expanding this WorkerBinder API given is going to be deprecated soon in favor of RibCoroutineWorker.

I was thinking though to add a migration mechanism to change the WorkerBinderDispatcher to other than Unconfined. In that manner we can migrate internally all worker binding outside Ui thread in a safe manner

FranAguilera:
Will be putting a new PR for migration API options

PR #596, opened 2023-06-13
Allow overriding default CoroutineDispatcher for WorkerBinder calls

FranAguilera:
**Description**:
WorkerBinder/Worker is going away in favor of RibCoroutineWorker. Still we would like to provide a mechanism to change the default CoroutineDispatcher being used at WorkerBinder.bind calls to be other than Unconfined. 

This is given that the effective threading for binding those calls happens on the main thread, potentially leading to Jank, ANR, freezes, etc.

**Related issue(s)**:

Current Workers being bound on Main thread. 

**Verification**:

- Verified on rib-workers demo app
image

- Existing JUnit tests

### DIFF
--- a/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
+++ b/android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
@@ -55,6 +55,10 @@ public object WorkerBinder {
 
   private var workerBinderListenerWeakRef: WeakReference<WorkerBinderListener>? = null
 
+  private var workerBinderDispatcher: WorkerBinderDispatcher = WorkerBinderDispatcher {
+    RibDispatchers.Unconfined
+  }
+
   /**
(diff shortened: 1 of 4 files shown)


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -59,28 +63,37 @@ public object WorkerBinder {
...
+  /**
+   * WARNING: Use with caution
+   *
+   * Changes the default WorkerBinderDispatcher used for all WorkerBinder calls. When not updated
+   * will rely on [RibDispatcher.Unconfined] for previous backward compatibility.
+   *
+   * To be called at the earliest point of your application
+   */
+  @JvmStatic
+  public fun updateDispatcher(workerBinderDispatcher: WorkerBinderDispatcher) {
```
FranAguilera, line 75: offline sync. setWorkerDefaultDispatcher

tyvsmith (approved):
Thanks for addressing the changes we discussed.

PR #599, opened 2023-06-14
Expose ribActionEvents stream

FranAguilera:
**Description**:
Expose `ribActionEvents` stream that emits state of each RIB "action" (e.g didBecomeActive, willLoad) For example:

```
Event to be emitted just before didBecomeActive
Interactor.didBecomeActive internal call
Even to be emitted after didBecomeActive call is completed
```
One example application of consuming these events is monitoring binding duration of Workers as shown in `ApplicationLevelWorkerLogger` from rib-workers demo app
...

_**WARNING: [RibEvents.enableRibActionEmissions] must be called at the earliest app scope before subscribing/collecting ribAction events. This is intentional in order to avoid unnecessary object creations for each RIB action in case no one is listening**_ 

**Breaking changes**:

| Change  | Notes |
| ------------- | ------------- |
|  `WorkerBinder.initializeMonitoring` | Removed. Moving forward can use new flow API |
| `WorkerBinderListener` | Removed. Moving forward can use new flow API |
| `RibEvents` | Converted to Object, no need for Singleton from now on |
| `RibEvent` | Named to `RibRouterEvent` |
| `events` within `RibEvents` | Renamed `events` to `routerEvents` |

**Tests**
- JUnit tests
- Manual validation on internal repo
- Demo on rib-workers sample app 
image

**Related issue(s)**:

Provide a new RibActionEvent stream that will allow for monitoring/logging options for all RIB components (Interactor, Presenter, Router, Worker)


FranAguilera:
Will be adding tests and cleaning up a bit more in a bit

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/RibEventLogger.kt
```diff
@@ -0,0 +1,45 @@
...
+ * Sample of consuming [ribEventDataFlow] possibilities.
+ * 1. Can pipe Interactor/Router/Presenter/Worker information to backend
+ * 2. Could report expensive workers on Ui thread and crash on Debug builds for early detection
+ * 3. More tailored aggregation if needed.
+ *
+ * IMPORTANT: Given logic of [onBindCompleted] will be running upon each Worker onStart/onStop.
+ * Added logic should guaranteed that we are not impacting app performance by having expensive logic
+ * within its implementation.
+ */
+class RibEventLogger {
```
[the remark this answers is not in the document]
FranAguilera: Oh yes, good catch! I still need to clean up this a bit but will update soon with tests as well

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -16,28 +16,82 @@
...
+   */
+  internal fun emitRibEventDuration(
+    ribClass: KClass<*>,
+    ribComponentType: RibComponentType,
+    ribEventType: RibEventType,
+    totalBindingDurationMilli: Long,
+  ) {
+    val ribClassName = ribClass.qualifiedName
 
-    @JvmStatic public fun getInstance(): RibEvents = instance
```
FranAguilera, line 41: @JavaOnly annotation to keep singleton

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -16,28 +16,82 @@
...
+  private val mutableRibDurationEvents =
+    MutableSharedFlow<RibEventDurationData>(0, 1, BufferOverflow.DROP_OLDEST)
+
+  @JvmStatic
+  public val routerEvents: Observable<RibRouterEvent>
+    get() = mutableRouterEvents.asObservable()
+
+  @JvmStatic
+  public val ribDurationEvents: Observable<RibEventDurationData>
+    get() = mutableRibDurationEvents.asObservable()
```
FranAguilera, line 37: remove get()

psteiger:
As discussed offline, we could have a DSL like:
```kotlin
inline fun record(startEvent: StartEvent, endEvent: EndEvent, block: () -> Unit) {
  // emit start event
  block()
  // emit end event
}
```

And in interactors (e.g.) we could have
```kotlin
fun didBecomeActive() {
  record(Interactor.Start, Interactor.Stop) {
      // code for didBecomeActive
  }
}
```

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: Actually, decided to remove this all-together and to replace existing usages moving forward

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
```diff
@@ -0,0 +1,78 @@
...
+      RibEvents.ribActionEvents
+        .filter { it.ribComponentType == RibComponentType.DEPRECATED_WORKER }
+        .asFlow()
+        .collect { it.logWorkerDuration() }
+    }
+  }
+
+  private fun RibActionInfo.logWorkerDuration() {
+    val ribComponentKey = this.className
+    if (ribActionState == RibActionState.STARTED && !ribComponentKey.isClassNameInMap()) {
```
FranAguilera, line 57: Measuring this locally on a low end device didn't see any performance implications. Either way, will validate with internal repo as well

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
```diff
@@ -157,9 +165,16 @@ protected constructor(
           .handleNonFatalWarning("A RIB tried to detach a child that was never attached", null)
       }
     }
-    childRouter.dispatchDetach()
+
+    triggerRibActionAndEmitEvents(
+      { childRouter.dispatchDetach() },
```
psteiger, line 170: This would probably be better as the leading parameter, so we can use trailing lambda syntax

FranAguilera: good idea, will update

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,24 +20,92 @@ import kotlinx.coroutines.channels.BufferOverflow
...
 
-public class RibEvents private constructor() {
+public object RibEvents {
 
-  private val _events = MutableSharedFlow<RibEvent>(0, 1, BufferOverflow.DROP_OLDEST)
-  public val events: Observable<RibEvent> = _events.asObservable()
+  private val mutableRouterEvents =
+    MutableSharedFlow<RibRouterEvent>(0, 1, BufferOverflow.DROP_OLDEST)
+  private val mutableRibDurationEvents =
+    MutableSharedFlow<RibActionInfo>(0, 1, BufferOverflow.DROP_OLDEST)
```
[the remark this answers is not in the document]
FranAguilera: Yes, I'm sort of modeling this in the same manner as previous rib "router" event. We can revisit later if issues down the road

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
```diff
@@ -0,0 +1,76 @@
...
+ * Interactor/Router/Presenter/Worker ATTACH/DETACH, the added logic should guaranteed that we are
+ * not impacting app performance
+ */
+object ApplicationLevelWorkerLogger {
+  private const val LOG_TAG = "RibEventLogger"
+
+  private val concurrentHashMap: ConcurrentHashMap<String, Long> = ConcurrentHashMap()
+
+  fun start() {
+    MainScope().launch(RibDispatchers.Default) {
```
psteiger, line 46: Probably better to use `GlobalScope`? Shows the intent of "running forever" better.

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
```diff
@@ -0,0 +1,76 @@
...
+ * 3. More tailored aggregation if needed.
+ *
+ * IMPORTANT: Given logic at [logWorkerDuration] will be running upon
+ * Interactor/Router/Presenter/Worker ATTACH/DETACH, the added logic should guaranteed that we are
+ * not impacting app performance
+ */
+object ApplicationLevelWorkerLogger {
+  private const val LOG_TAG = "RibEventLogger"
+
+  private val concurrentHashMap: ConcurrentHashMap<String, Long> = ConcurrentHashMap()
```
psteiger, line 43: Maybe set the type to just `Map` (while still keeping the implementation as ConcurrentHashMap)

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  private fun RibActionInfo.logWorkerDuration() {
+    val ribComponentKey = this.className
+    if (ribActionState == RibActionState.STARTED && !ribComponentKey.isClassNameInMap()) {
+      concurrentHashMap[ribComponentKey] = currentTimeMillis()
+    }
+
+    if (ribActionState == RibActionState.COMPLETED && ribComponentKey.isClassNameInMap()) {
```
psteiger, line 60: Does it make sense to use `else if` instead?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,24 +20,96 @@ import kotlinx.coroutines.channels.BufferOverflow
...
+  /**
+   * Calls related RIB action (e.g. didBecomeActive) and emits emission of ATTACHED/DETACHED events
+   * for each RIB component.
+   *
+   * @param ribAction The related RIB action type. e.g. didBecomeActive, willLoad, etc
+   * @param ribCallerClassType Related RIB component class type
+   * @param ribComponentType The RIB component type (e.g. Interactor, Router, Presenter, Worker)
+   * @param ribEventType RIB event type (e.g. ATTACH/DETACH)
+   */
+  internal fun <T : Any> triggerRibActionAndEmitEvents(
```
psteiger, line 55: If it's possible we could consider making it inline to save on the lambda stack allocation (but paying in binary size output).

Just a thought.

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
FranAguilera: totally agree :)

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
FranAguilera: yup

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: good idea, will check it out

android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/logger/ApplicationLevelWorkerLogger.kt
FranAguilera: good idea

psteiger (approved):

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibComponentType.kt
```diff
@@ -13,8 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.uber.rib.workers.root.main.workers.monitoring
+package com.uber.rib.core
 
-fun interface BackendReporter {
-  fun report(genericMessage: String)
+public enum class RibComponentType {
```
tyvsmith, line 18: Regular Worker?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,24 +20,96 @@ import kotlinx.coroutines.channels.BufferOverflow
...
+   * Calls related RIB action (e.g. didBecomeActive) and emits emission of ATTACHED/DETACHED events
+   * for each RIB component.
+   *
+   * @param ribAction The related RIB action type. e.g. didBecomeActive, willLoad, etc
+   * @param ribCallerClassType Related RIB component class type
+   * @param ribComponentType The RIB component type (e.g. Interactor, Router, Presenter, Worker)
+   * @param ribEventType RIB event type (e.g. ATTACH/DETACH)
+   */
+  internal inline fun <T : Any> triggerRibActionAndEmitEvents(
+    ribCallerClassType: T,
```
tyvsmith, line 56: Can we lock this down more than T? With the use of lambda's, this, it, and T here it seems like the risk is high that we end up not tracking the correct class we're expecting.

Is there a common interface we can use here? If not, should we have one?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibComponentType.kt
FranAguilera: We may need to tweak a bit more RibCoroutine worker in order to provide accurate monitoring as mentioned by @psteiger earlier 

Given there are barely any users internally of RibCoroutineWorker, I'm not concerned to leave these changes to be implemented on a separate PR

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: Good idea, let me explore

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,24 +20,96 @@ import kotlinx.coroutines.channels.BufferOverflow
...
 
   /**
    * @param eventType [RibEventType]
    * @param child [Router]
    * @param parent [Router] and null for the root ribs that are directly attached to
    *   RibActivity/Fragment
    */
-  public fun emitEvent(eventType: RibEventType, child: Router<*>, parent: Router<*>?) {
-    _events.tryEmit(RibEvent(eventType, child, parent))
+  public fun emitRouterEvent(eventType: RibEventType, child: Router<*>, parent: Router<*>?) {
```
tyvsmith, line 42: 1. Does this need to remain public?
2. If we're changing the API already, maybe best to make @JvmStatic so it's consistent with other accessors in Java?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,24 +20,96 @@ import kotlinx.coroutines.channels.BufferOverflow
...
+   * @param ribEventType RIB event type (e.g. ATTACH/DETACH)
+   */
+  internal inline fun <T : Any> triggerRibActionAndEmitEvents(
+    ribCallerClassType: T,
+    ribComponentType: RibComponentType,
+    ribEventType: RibEventType,
+    ribAction: () -> Unit,
+  ) {
+    val ribClassName = ribCallerClassType.asQualifiedName()
+    ribClassName?.emitRibEventAction(ribComponentType, ribEventType, RibActionState.STARTED)
```
tyvsmith, line 62: This function extension doesn't make sense. Just take a string class name as a param on the method.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,24 +20,96 @@ import kotlinx.coroutines.channels.BufferOverflow
...
+  ) {
+    val ribActionInfo =
+      RibActionInfo(
+        this,
+        ribComponentType,
+        ribEventType,
+        ribActionState,
+        Thread.currentThread().name,
+      )
+    mutableRibDurationEvents.tryEmit(ribActionInfo)
```
tyvsmith, line 89: Maybe we want to take an explicit flag that enables event callbacks (or check for subcribers) to avoid creating objects if no one is listening.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: 1. Unfortunately yes, this is used within RibActivity living in `rib-android` module. Also, in our internal repo is used in multiple places. I can create an issue and follow up on this one

2. Good idea, will apply

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: Added a "RibComponent" contract to this common components, still thinking a better naming for this

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: Added a factory `RibEvents.startCapturingRibActionInfo()` to be called explicitly before any ribActionEvents collection

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -48,39 +60,48 @@ public object RibEvents {
...
-   * @param ribCallerClassType Related RIB component class type
+   * @param ribComponent Related RIB component
    * @param ribComponentType The RIB component type (e.g. Interactor, Router, Presenter, Worker)
    * @param ribEventType RIB event type (e.g. ATTACH/DETACH)
    */
-  internal inline fun <T : Any> triggerRibActionAndEmitEvents(
-    ribCallerClassType: T,
+  internal inline fun triggerRibActionAndEmitEvents(
+    ribComponent: RibComponent,
     ribComponentType: RibComponentType,
```
FranAguilera, line 69: I would like to specify a property on the new RibComponent to be overriden, unfortunately given that Worker and RibCoroutineWorker are interfaces we cannot restrict visibility of those

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -33,12 +33,24 @@ public object RibEvents {
...
   public val ribActionEvents: Observable<RibActionInfo> = mutableRibDurationEvents.asObservable()
 
+  private var allowRibActionEmissions = false
+
+  /**
+   * To be called before start observing/collecting on [ribActionEvents] (usually at your earliest
+   * application point)
+   */
+  @JvmStatic
+  public fun startCapturingRibActionInfo() {
```
tyvsmith: Do we need a disable too? For testing?

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
```diff
@@ -34,7 +34,7 @@ protected constructor(
   public open val interactor: I,
   private val ribRefWatcher: RibRefWatcher,
   private val mainThread: Thread,
-) {
+) : RibComponent {
```
tyvsmith, line 37: Probably too generic for a name for this set of callbacks.  How about something more explicit around emitting callbacks around lifecycles, like RibEventEmitter?

FranAguilera: Good idea, I like this better

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
FranAguilera: I agree that will be useful, will add shortly

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
FranAguilera: Decided to go with `RibActionEmitter` to avoid confusion with existing `RibEventType`

tyvsmith (approved):

PR #604, opened 2023-07-25
Add `WorkerBinder.bind` overloads that take in an `Iterable`.

psteiger:
`List` is not really needed and we restrict the API  unnecessarily: all we need is an `Iterable`.

For keeping binary compatibility, we also keep the overloads taking in a `List`.


tyvsmith (approved):

Gdakgdak22 (commented):
Revisando

PR #606, opened 2023-08-07
Set view tree owners

jbarr21:
Set view tree owners in `RibActivity#onCreate`. Since it adds the root RIB's view to the `android.R.id.content` view using `Activity#addView(View)` instead of calling [AppCompatActivity#setContentView(View)](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:appcompat/appcompat/src/main/java/androidx/appcompat/app/AppCompatActivity.java;l=202?q=appcompatactivity&ss=androidx) which already handles this.

### DIFF
--- a/android/libraries/rib-android/src/test/kotlin/com/uber/rib/core/RibActivityTest.kt
+++ b/android/libraries/rib-android/src/test/kotlin/com/uber/rib/core/RibActivityTest.kt
@@ -20,6 +20,7 @@ import android.content.Intent
 import android.view.View
 import android.view.ViewGroup
 import android.widget.FrameLayout
+import androidx.lifecycle.findViewTreeLifecycleOwner
 import com.google.common.truth.Truth.assertThat
 import com.uber.autodispose.AutoDispose
 import com.uber.autodispose.lifecycle.LifecycleEndedException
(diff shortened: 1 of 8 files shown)


android/libraries/rib-android/src/test/kotlin/com/uber/rib/core/RibActivityTest.kt
```diff
@@ -189,6 +190,17 @@ class RibActivityTest {
...
 
+  @Test
+  fun onCreate_setsViewTreeOwners() {
+    val activity = Robolectric.buildActivity(EmptyActivity::class.java).create(null).get()
+    assertThat(activity.window.decorView.findViewTreeLifecycleOwner()).isNotNull()
+
+    val contentView = activity.findViewById<FrameLayout>(android.R.id.content)
+    val rootView = View(RuntimeEnvironment.application)
+    contentView.addView(rootView)
+    assertThat(rootView.findViewTreeLifecycleOwner()).isNotNull()
```
psteiger, line 201: Can we assert for `ViewTreeSavedStateRegistry` and `ViewTreeViewModelStoreOwner` as well?

psteiger (changes requested):

android/libraries/rib-android/src/test/kotlin/com/uber/rib/core/RibActivityTest.kt
jbarr21: Added more assertions in 72b5415

psteiger (approved):

PR #607, opened 2023-08-10
Add RibCoroutineWorker.bind that receives multiple workers

FranAguilera:
**Description**:

Adding API to allow binding multiple RibCoroutineWorkers at once (similarly to WorkerBinder.bind added in https://github.com/uber/RIBs/pull/604) 

**Related issue(s)**:

Avoid boilerplate to bind multiple workers. This is very common use case internally.

**Verification**:

Verified via Rib worker sample app


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibCoroutineWorker.kt
```diff
@@ -107,6 +107,17 @@ public fun CoroutineScope.bind(
   return BindWorkerHandleImpl(bindJob, unbindJob)
 }
 
+/** Binds a list of [worker]'s in a scope that is a child of the [CoroutineScope] receiver. */
```
psteiger, line 110:
```suggestion
/** Binds a list of [workers] in a scope that is a child of the [CoroutineScope] receiver. */
```

psteiger (changes requested):

tyvsmith (approved):

psteiger (approved):

PR #609, opened 2023-08-22
Update coroutines 1.7.3

tyvsmith:
Updating to 1.7.3 and removing unneeded experimental API annotations
[Changelog](https://github.com/Kotlin/kotlinx.coroutines/releases)


FranAguilera (approved):

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/RibCoroutinesRule.kt
```diff
@@ -15,24 +15,25 @@
...
 /**
  * RibCoroutinesRule is a Junit TestRule to act as a managed TestCoroutineScope in test and to
  * facilitate install and cleanup of Test Dispatchers
  */
-@ExperimentalCoroutinesApi
 public class RibCoroutinesRule(
   public val ribDispatchers: TestRibDispatchers = TestRibDispatchers(),
 ) : TestWatcher() {
 
+  private var originalDeprecatedWorkerDispatcher = RibCoroutinesConfig.deprecatedWorkerDispatcher
```
psteiger, line 29: Should we set this property in `starting` instead? Wonder if there's any change that:
1. `deprecatedWorkerDispatcher` is set to `A` (class instantiated)
2. Unrelated code sets `deprecatedWorkerDispatcher` to `B`
3. `starting` sets to `ribDispatchers.Unconfined`
4. `finished` sets to `A` but should be `B`

Probably not.

psteiger (approved):

psteiger (approved):

android/libraries/rib-coroutines-test/src/main/kotlin/com/uber/rib/core/RibCoroutinesRule.kt
tyvsmith: Yeah, adjusted to set in starting.

PR #612, opened 2023-08-23
proposal to detach children before self

davissuber:
This pull request stems from this issue and attempt to swap the detachment order between self and any attached children to avoid a small window in which the child is active and parent is not.


tyvsmith:
Has this been verified in the Uber monorepo? Tests passed and apps worked?

davissuber:
> Has this been verified in the Uber monorepo? Tests passed and apps worked?

No, not yet - was creating this pull request to collect information on whether to do this at all (and if yes, gather information on what would be considered a sufficient test)

tyvsmith:
Suggest a first pass with a custom built dependency to understand the scope of the migration

davissuber:
Unfortunately running into some crashes with the first pass. Some existing code expect the parent's `willDetach` to be called before the child's `dispatchDetach` (e.g. `detachChild`).

Make the call order "will detach" -> "detach children" -> "detach self" would solve that problem, but that seems like overfitting?

android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/InteractorAndRouterTest.kt
```diff
@@ -97,16 +97,23 @@ class InteractorAndRouterTest {
...
+      }
+    val selfInteractorDetachIndex =
+      ribActionInfoValues.indexOfFirst {
+        it.ribEventType == RibEventType.DETACHED &&
+          it.ribActionEmitterType == RibActionEmitterType.INTERACTOR &&
+          it.ribActionState == RibActionState.COMPLETED
+      }
+    assertThat(selfInteractorDetachIndex).isGreaterThan(-1)
+    assertThat(childRouterDetachIndex).isGreaterThan(-1)
+    assertThat(selfInteractorDetachIndex > childRouterDetachIndex).isTrue()
```
davissuber, line 116: Since we are detaching child before parent now, the router detach is no longer the final event.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Router.kt
```diff
@@ -206,11 +206,11 @@ protected constructor(
   public open fun dispatchDetach() {
     checkForMainThread()
 
-    interactorGeneric.dispatchDetach()
     willDetach()
```
psteiger, line 209: Should the router `willDetach` come *after* interactor detach? IMO this should be the last line. What issues did you encounter with this (if any)?

psteiger: Upon second though, maybe calling willDetach before detaching children and self makes some sense after all. "Will" implies some kind of preparation for something that is about to happen.

davissuber: What I had noticed was that some existing RIBs in Uber's app are overriding the `willDetach()` (and sometimes even call `detachChild()` themselves e.g. BikeHomeRouter) and/or do other sorts of clean up stuff that needs to happen before `detachChild()` is called by Router.

I think those seemingly ani-pattern-ish usage may be symptoms of the fact that we are detaching parent before child.

psteiger: I think you are right.

I've been thinking about this, there's nothing super wrong with calling willDetach before everything else. It can be understood as a signal that router X is **starting** its detach process, which now will _begin_ with detaching children (instead with detaching self). That makes sense.

psteiger (approved):
If smoke test went fine, I think this is a welcome change. We should consider pushing for iOS changing it as well, for symmetry .

PR #614, opened 2023-09-06
Improve XRay implementation

lucasnlm:
**Description**:
This PR adds `setup()`, `XRayConfig` to RIBs XRay to allow setup it without rely on the existing `toggle()` function.

** Problem it fixes **
The current `toggle()` function may be problematic because it changes a static boolean. Example: depending on where XRay is enabled or if it's as debug in a RIB used twice, it will enabled and disable immediately.

`XRayConfig` adds the ability to enable XRay without change the Views alpha. The alpha change may be problematic when we have many RIBs attached due to the overlapping.

### DIFF
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/ViewRouter.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/ViewRouter.kt
@@ -34,7 +34,7 @@ abstract class ViewRouter<V : View, I : Interactor<*, *>> : Router<I> {
   ) : super(interactor, component) {
     this.view = view
     if (XRay.isEnabled()) {
-      XRay.apply(this, view)
+      XRay.apply(getName(), view)
     }
   }
 
@@ -44,7 +44,7 @@ abstract class ViewRouter<V : View, I : Interactor<*, *>> : Router<I> {
   ) : super(null, interactor, RibRefWatcher.getInstance(), getMainThread()) {
     this.view = view
     if (XRay.isEnabled()) {
-      XRay.apply(this, view)
+      XRay.apply(getName(), view)
     }
   }
 
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
@@ -23,58 +23,60 @@ import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import android.view.Gravity
 import android.view.View
+import androidx.annotation.VisibleForTesting
 
 /** Utility class that shows riblets name in its background. */
 class XRay private constructor() {
-  private var isEnabled = false
-  private var textPaint: Paint? = null
-  private fun writeOnBitmap(bitmap: Bitmap, text: String) {
-    val canvas = Canvas(bitmap)
-    val textPaint = getTextPaint()
-    val xStartPoint = (bitmap.width - textPaint.measureText(text)) / 2f
-    val yStartPoint = bitmap.height / 2f
-    canvas.drawText(text, xStartPoint, yStartPoint, textPaint)
+  private var config = XRayConfig()
+  private val textPaint: Paint by lazy {
+    Paint().apply {
+      textSize = TEXT_SIZE.toFloat()
+      color = TEXT_COLOR
+    }
   }
 
-  private fun getTextPaint(): Paint {
-    if (textPaint == null) {
-      textPaint =
-        Paint().apply {
-          textSize = TEXT_SIZE.toFloat()
-          color = TEXT_COLOR
-        }
+  private fun writeOnBitmap(bitmap: Bitmap, text: String) {
+    Canvas(bitmap).run {
+      val xStartPoint = (bitmap.width - textPaint.measureText(text)) / 2f
+      val yStartPoint = bitmap.height / 2f
+      drawText(text, xStartPoint, yStartPoint, textPaint)
     }
-    return textPaint!!
   }
 
   companion object {
     private val INSTANCE = XRay()
     private const val FRAME_WIDTH = 500
     private const val FRAME_HEIGHT = 150
-    private const val XRAY_ALFA = 0.9f
+    private const val XRAY_ALPHA = 0.9f
     private const val TEXT_SIZE = 30
     private const val TEXT_COLOR = Color.RED
 
-    /** Toggles state of XRay. */
+    /** Setup XRay using a [XRayConfig] */
     @JvmStatic
-    fun toggle() {
-      INSTANCE.isEnabled = !INSTANCE.isEnabled
+    public fun setup(config: XRayConfig) {
+        INSTANCE.config = config
+    }
+
+    /** Toggles state of XRay. */
+    public fun toggle() {
+      val config = INSTANCE.config
+      setup(config.copy(enabled = !config.enabled))
     }
 
     /** @return `true` if XRay is enabled, `false` otherwise. */
     @JvmStatic
-    fun isEnabled(): Boolean {
-      return INSTANCE.isEnabled
+    public fun isEnabled(): Boolean {
+      return INSTANCE.config.enabled
     }
 
     /**
      * Puts [ViewBuilder]s riblet name in the background of the [View]
      *
-     * @param viewRouter a [ViewRouter] which riblets name should be written.
+     * @param routerName the riblets name to be written.
      * @param view a [View] to put the name behind.
      */
     @JvmStatic
-    fun apply(viewRouter: ViewRouter<*, *>, view: View) {
+    internal fun apply(routerName: String, view: View) {
       val oldBackground = view.background
       val bitmap: Bitmap =
         if (oldBackground != null) {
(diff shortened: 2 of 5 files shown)


android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/ViewRouter.kt
```diff
@@ -34,7 +34,7 @@ abstract class ViewRouter<V : View, I : Interactor<*, *>> : Router<I> {
   ) : super(interactor, component) {
     this.view = view
     if (XRay.isEnabled()) {
-      XRay.apply(this, view)
+      XRay.apply(getName(), view)
```
lucasnlm, line 37: Change this because it was leaking a `this` reference to a static function.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
```diff
@@ -23,70 +23,76 @@ import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import android.view.Gravity
 import android.view.View
+import androidx.annotation.VisibleForTesting
 
 /** Utility class that shows riblets name in its background. */
 class XRay private constructor() {
-  private var isEnabled = false
-  private var textPaint: Paint? = null
```
lucasnlm, line 30: Changed `textPaint` to a lazy `val`.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
```diff
@@ -23,70 +23,76 @@ import android.graphics.drawable.BitmapDrawable
...
+    /** Setup XRay using a [XRayConfig] */
     @JvmStatic
-    fun toggle() {
-      INSTANCE.isEnabled = !INSTANCE.isEnabled
+    public fun setup(config: XRayConfig) {
+        INSTANCE.config = config
+    }
+
+    /** Toggles state of XRay. */
+    public fun toggle() {
```
lucasnlm, line 61: `toggle` is still here to keep compatibility.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
```diff
@@ -23,70 +23,76 @@ import android.graphics.drawable.BitmapDrawable
...
     /**
      * Puts [ViewBuilder]s riblet name in the background of the [View]
      *
-     * @param viewRouter a [ViewRouter] which riblets name should be written.
+     * @param routerName the riblets name to be written.
      * @param view a [View] to put the name behind.
      */
     @JvmStatic
-    fun apply(viewRouter: ViewRouter<*, *>, view: View) {
+    internal fun apply(routerName: String, view: View) {
```
lucasnlm, line 79: `viewRouter` was used only to get its name.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
psteiger: If it's all static info, why not immediately initialize it?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
psteiger: Good change

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
```diff
@@ -23,70 +23,74 @@ import android.graphics.drawable.BitmapDrawable
...
+    Canvas(bitmap).run {
+      val xStartPoint = (bitmap.width - textPaint.measureText(text)) / 2f
+      val yStartPoint = bitmap.height / 2f
+      drawText(text, xStartPoint, yStartPoint, textPaint)
     }
-    return textPaint!!
   }
 
   companion object {
     private val INSTANCE = XRay()
```
psteiger, line 47: Consider changing the `class` to an `object`, if it's supposed to be a singleton after all. Then we can get rid of `companion object`

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
psteiger: If so, I suggest we deprecate it.
Is there a substituting method?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRayConfig.kt
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Configuration for XRay.
+ *
+ * @property enabled `true` to enable XRay. By default it is disabled.
+ * @property alphaEnabled `true` to enable alpha changes when XRay is enabled.
+ */
+public data class XRayConfig(
+  val enabled: Boolean = false,
+  val alphaEnabled: Boolean = true,
+)
```
psteiger, line 27: This is so simple IMO it should live on `XRay.kt` file at the top level. No need for a new file

psteiger (changes requested):

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
```diff
@@ -23,70 +23,74 @@ import android.graphics.drawable.BitmapDrawable
...
+    private const val XRAY_ALPHA = 0.9f
     private const val TEXT_SIZE = 30
     private const val TEXT_COLOR = Color.RED
 
-    /** Toggles state of XRay. */
+    /** Setup XRay using a [XRayConfig] */
     @JvmStatic
-    fun toggle() {
-      INSTANCE.isEnabled = !INSTANCE.isEnabled
+    public fun setup(config: XRayConfig) {
```
psteiger, line 56: Do we need thread-safety? If so, we must instead provide a way to atomically "read and write" -- an `.update { }` function, as below:

```
INSTANCE = AtomicReference<XRayConfig>(XRayConfig())

inline fun update(function: (current: T) -> T) {
  INSTANCE.getAndUpdate(function)
}
```

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
lucasnlm: `Paint` is a big object. I rather initialize it only if needed. WDYT?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
lucasnlm: Makes sense! I will do that.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
lucasnlm: I think we don´t need thread-safety. At least I never had thread issues with XRay.
Is it ok for you if we let it as is?

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
lucasnlm: Done.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRayConfig.kt
lucasnlm: Done

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
lucasnlm: BTW, XRay is called in every `ViewRouter` constructor. So, I will initialize `Paint` every time. Even if we don't want to use XRay.

psteiger: @lucasnlm but this is a singleton object, right? So it will really only initialize once. There's only one instance of the `XRay` class, or am I misunderstanding it?

psteiger: "I rather initialize it only if needed. WDYT?"
Ok this makes sense.

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/XRay.kt
psteiger: Yes it's ok, was just wondering.

psteiger (approved):
I'd probably not make Paint lazy, as the XRay object itself will only be initialized first time it is called anyway, so by lazy in its property for those purposes feels redundant 

Stamped though

psteiger:
@lucasnlm please build a version of RIBs with your change and make sure it works as expected using our apps, if so we can go ahead with landing

lucasnlm:
@psteiger How can I do that?

psteiger:
@lucasnlm I tagged on in internal comms with smoke test instructions

lucasnlm:
> @lucasnlm I tagged on in internal comms with smoke test instructions

I hope to test it this week. Thanks for the reminder.

PR #615, opened 2023-09-12
Set view tree owners in parent class

jbarr21:
A followup to #606 which moves the logic into the superclass so all `CoreAppCompatActivity` subclasses benefit from the updated logic

### DIFF
--- a/android/libraries/rib-android-core/src/main/kotlin/com/uber/rib/core/CoreAppCompatActivity.kt
+++ b/android/libraries/rib-android-core/src/main/kotlin/com/uber/rib/core/CoreAppCompatActivity.kt
@@ -20,6 +20,9 @@ import android.os.Bundle
 import androidx.annotation.CallSuper
 import androidx.annotation.IntRange
 import androidx.appcompat.app.AppCompatActivity
+import androidx.lifecycle.ViewTreeLifecycleOwner
+import androidx.lifecycle.ViewTreeViewModelStoreOwner
+import androidx.savedstate.ViewTreeSavedStateRegistryOwner
 
 /** Core Support v7 AppCompat Activity. */
 abstract class CoreAppCompatActivity : AppCompatActivity() {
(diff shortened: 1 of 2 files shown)


android/libraries/rib-android-core/src/main/kotlin/com/uber/rib/core/CoreAppCompatActivity.kt
```diff
@@ -81,4 +85,16 @@ abstract class CoreAppCompatActivity : AppCompatActivity() {
...
+  /**
+   * [RibActivity] must call this since it does not use [ComponentActivity.setContentView] which
+   * already handles this.
+   */
+  private fun initViewTreeOwners() {
+    // Set the view tree owners before setting the content view so that the inflation process
+    // and attach listeners will see them already present
+    ViewTreeLifecycleOwner.set(window.decorView, this)
+    ViewTreeViewModelStoreOwner.set(window.decorView, this)
+    ViewTreeSavedStateRegistryOwner.set(window.decorView, this)
```
psteiger, line 98: I guess there are now `setViewTreeXyz` extensions that don't need to take in `this`

psteiger (approved):

android/libraries/rib-android-core/src/main/kotlin/com/uber/rib/core/CoreAppCompatActivity.kt
jbarr21: this artifact was already there & the ext fun weren't visible so i stuck w/ this

idanakav (approved):

android/libraries/rib-android-core/src/main/kotlin/com/uber/rib/core/CoreAppCompatActivity.kt
```diff
@@ -81,4 +85,16 @@ abstract class CoreAppCompatActivity : AppCompatActivity() {
     super.onActivityResult(requestCode, resultCode, data)
     activityDelegate?.onActivityResult(this, requestCode, resultCode, data)
   }
+
+  /**
+   * [RibActivity] must call this since it does not use [ComponentActivity.setContentView] which
```
psteiger, line 90: What happens for activities that extend from this and call setContentView? I imagine lifecycle owners get overwritten and that's ok

jbarr21: overwritten & ok

PR #617, opened 2023-09-20
Simplify some verifications in Interactor

lucasnlm:
**Description**:
This PR simplify some verifications in Interactor that can be done in other places or are no required anymore.

**Related issue(s)**:
NA

### DIFF
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
@@ -96,10 +96,9 @@ abstract class RibActivity :
     initViewTreeOwners()
     val rootViewGroup = findViewById<ViewGroup>(android.R.id.content)
     _lifecycleFlow.tryEmit(createOnCreateEvent(savedInstanceState))
-    val wrappedBundle: Bundle? =
-      if (savedInstanceState != null) Bundle(savedInstanceState) else null
     router = createRouter(rootViewGroup)
     router?.let {
+      val wrappedBundle: Bundle? = savedInstanceState?.let(::Bundle)
       it.dispatchAttach(wrappedBundle)
       rootViewGroup.addView(it.view)
       RibEvents.emitRouterEvent(RibEventType.ATTACHED, it, null)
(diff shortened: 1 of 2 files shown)


android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -96,10 +96,9 @@ abstract class RibActivity :
     initViewTreeOwners()
     val rootViewGroup = findViewById<ViewGroup>(android.R.id.content)
     _lifecycleFlow.tryEmit(createOnCreateEvent(savedInstanceState))
-    val wrappedBundle: Bundle? =
-      if (savedInstanceState != null) Bundle(savedInstanceState) else null
     router = createRouter(rootViewGroup)
     router?.let {
+      val wrappedBundle: Bundle? = savedInstanceState?.let(::Bundle)
```
lucasnlm, line 101: `wrappedBundle` is used only if router is not null. So, moved to this let.

android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -150,26 +150,15 @@ public abstract class Interactor<P : Any, R : Router<*>>() : InteractorType, Rib
...
-          actualPresenter
-        } else {
-          injectedPresenter
-        }
-      } catch (e: UninitializedPropertyAccessException) {
-        actualPresenter
-      }
-    checkNotNull(presenter) { "Attempting to get interactor's presenter before being set." }
-    return presenter
+    return runCatching { actualPresenter ?: injectedPresenter }
```
psteiger, line 159: `runCatching` is designed for wrapping some exception for later rethrowing/handling. It should almost never be used as a catch-all that just swallows the generic `Throwable` (it catches everything). 

I suggest we maintain catching only the specific exception, though the code for sure could be improved in style somehow.

See [Encapsulate successful or failed function execution](https://github.com/Kotlin/KEEP/blob/master/proposals/stdlib/result.md) for use cases and motivation of `runCatching`

psteiger (changes requested):

android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
psteiger: We could probably go further and
```
    createRouter(rootViewGroup).let { r ->
      router = r
      val wrappedBundle = savedInstanceState?.let(::Bundle)
      r.dispatchAttach(wrappedBundle)
      rootViewGroup.addView(r.view)
      RibEvents.emitRouterEvent(RibEventType.ATTACHED, r, null)
    }
```

To avoid the unnecessary null check (router will never be null at this point)

PR #618, opened 2023-09-21
Get rid of suppressions for "invisible_reference" and "invisible_member"

psteiger:
For accessing internal declarations from "friend" modules (e.g. access `rib-base` internals from `rib-test`), we are currently suppressing compiler errors with `@file:Suppress("invisible_reference", "invisible_member")`

A better approach is to:
1. Create an `internal` opt-in annotation.
2. Make the "accessible to friend modules" component `public`
3. Mark the (now public) component with the opt-in annotation.
4. Opt-in to the annotation from `build.gradle`.

Because the new annotation is `internal`, it cannot be normally accessed from external modules. But Gradle can see it if it's part of the source set.

Benefits:
1. We get rid of the hacky suppressions in our codebase.
2. By suppressing `invisible_reference` and `invisible_member` we gain access to *all* internal components from the other modules, and that's not ideal: we'd rather be explicit about what are the APIs that are visible for module friends, and what are truly internal for the module and should never be used even from module friends.
3. We make the internal visibility of those friend-modules APIs even stricter, since consumers from the general public now cannot just suppress `"invisible_reference"` and `"invisible_member"` to directly access the friend-module API -- they require explicit Opt-In. They would need to suppress invisible members AND opt-in.

### DIFF
--- a/android/libraries/rib-android/build.gradle.kts
+++ b/android/libraries/rib-android/build.gradle.kts
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2017. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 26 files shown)


psteiger:
Note: this is the same approach now being used in KotlinX libraries: https://github.com/Kotlin/kotlinx.serialization/pull/2444

FranAguilera (approved):

PR #534, opened 2023-01-02
Fix: podspec and tutorial1 Podfile version updated

ebubekirsezer:
**Description**:
Because of the low version of RxSwift and RxRelay in podspec file, pod install gives error. To resolve this issue in tutorial1 project i updated the podspec and Podfile

- Before
Screenshot 2023-01-03 at 00 01 03

- After
Screenshot 2023-01-03 at 00 01 13

**Related issue(s)**:
#312 
#389 
#354 
#388

PR #619, opened 2023-10-02
Fix Podfile versions for tutorials 2, 3, 4

elliot-360:
**Description**:
The main RIBs project currently uses version 6.5 of RxCocoa. Because of this, if you try to `pod install` during the tutorial, there will be a conflict because the tutorials currently use version 5.1 of RxCocoa.

See this PR: https://github.com/uber/RIBs/pull/534 that fixed the problem for tutorial 1. This PR does the same thing for tutorials 2, 3, and 4.


elliot-360:
Hi @rudro, would you mind taking a look at this? Thank you!

PR #620, opened 2023-10-03
Introduce `TestScope.test(RibCoroutineWorker)` test helper utility.

psteiger:
This helper utility is meant to be used in tests inside `runTest { }` blocks and should facilitate `RibCoroutineWorker` testing by automatically binding and unbinding the worker.

1. A test dispatcher is used to bind the worker (`StandardTestDispatcher`), so the test shall work even when `RibCoroutinesRule` is not used to override the `RibDispatchers.Default` dispatcher with a test dispatcher.
2. Worker is only bound in the scope of the lambda passed to `test(RibCoroutineWorker)`, hence tests never timeout with uncompleted coroutines because of a bound worker.

```
@Test fun test() = runTest {
  test(worker) {
    // Worker is bound. Make assertions
  }
  // Worker is unbound.
}
```

### DIFF
--- a/android/libraries/rib-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineWorker.kt
+++ b/android/libraries/rib-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineWorker.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2023. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.rib.core
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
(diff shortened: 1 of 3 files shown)


android/libraries/rib-test/src/main/kotlin/com/uber/rib/core/TestRibCoroutineWorker.kt
```diff
@@ -0,0 +1,53 @@
...
+ * will already be bound at the start of [testBody] lambda. If there are delays, calling
+ * [advanceTimeBy] or [advanceUntilIdle] at the start of [testBody] is needed to complete the
+ * binding.
+ *
+ * The same rationale applies for coroutines launched in the [CoroutineScope] parameter of
+ * [RibCoroutineWorker.onStart]: if there are no delays involved, coroutines will be run until idle
+ * or completed, otherwise, the aforementioned time advancing API must be used.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+public inline fun <T : RibCoroutineWorker> TestScope.test(
```
davissuber, line 40: Curious, why not place this function somewhere in `rib-coroutines-test` instead of here?

davissuber: Also, this file has the same name as the private class `TestRibCoroutineWorker` in `RibCouroutineWorkerTest.kt` (no immediate suggestion, just saying that I had to do a double-take.)

psteiger: > Curious, why not place this function somewhere in `rib-coroutines-test` instead of here?

Because `RibCoroutineWorker` is in `rib-base`, not `rib-coroutines`. 

> Also, this file has the same name as the private class TestRibCoroutineWorker in RibCouroutineWorkerTest.kt (no immediate suggestion, just saying that I had to do a double-take.)

Hm. Maybe it's reasonable to rename the private class to something like `TestWorker`

FranAguilera (approved):

PR #621, opened 2023-10-31
Refactor internal methods to avoid duplicate JVM signature

jbarr21:
`ViewRouter` extends from `Router` & both have a method: `internal fun onSaveInstanceStateInternal(outState: Bundle)`.
In order to avoid exposing this method to non-RIB-class subclasses, the method was duplicated each Router class w/ internal visibility. But, starting in K2 this fails with duplicate JVM signature, so this PR removes the `ViewRouter` copy & makes the `Router` copy public but requiring a core annotation.

### DIFF
--- a/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
+++ b/android/libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
@@ -104,7 +104,7 @@ public abstract class RibActivity :
   override fun onSaveInstanceState(outState: android.os.Bundle) {
     super.onSaveInstanceState(outState)
     _callbacksFlow.tryEmit(createOnSaveInstanceStateEvent(outState))
-    router?.dispatchSaveInstanceState(Bundle(outState))
+    router?.saveInstanceStateInternal(Bundle(outState))
       ?: throw NullPointerException("Router should not be null")
   }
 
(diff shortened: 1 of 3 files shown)


psteiger (changes requested):
I propose a different approach:

1. Completely kill `ViewRouter.saveInstanceStateInternal` -- it seems the only reason it existed was to force a friend-module visibility by duplicating the method name in a class in the module that will access it -- kinda hacky.
2. Make `Router.saveInstanceStateInternal` `public`, but annotate it with `@CoreFriendModuleApi`.

`RibActivity` now can access `Router.saveInstanceStateInternal` because `rib-android` is a core-friend module (opt-in from Gradle)

Now, external Kotlin consumers can't just `@OptIn(CoreFriendModuleApi::class)` because it is an internal annotation. So it remains inaccessible to them.

External Java consumers can access it, but that's the same issue with just `internal` visibility anyway -- it's public but mangled.

See https://github.com/uber/RIBs/pull/618

psteiger:
Let me know what you think of the suggested approach -- I think it's being pushed by JetBrains own libraries to handle the "internal to friend modules" visibility issue and I'm pushing for the same here.

jbarr21:
I like it, thanks for sharing. Definitely feels better, too

psteiger (approved):

PR #629, opened 2023-12-03
Support classes that are both `Worker` and `RibCoroutineWorker` in co…

psteiger:
…nverters.

`Worker.asRibCoroutineWorker` and `RibCoroutineWorker.asWorker` converters now return the instance unchanged if the class already implements the target interface.

This is important for correctness. For example:

```
class DualWorker : Worker, RibCoroutineWorker {
  fun onStart(lifecycle: WorkerScopeProvider) {
    doX()
  }

  suspend fun onStart(scope: CoroutineScope) {
    doY()
  }
}
```

Binding this `worker.asWorker()` should call `doX()`, not `doY()`, and binding this `worker.asRibCoroutineWorker()` should call `doY()`, not `doX()`.

### DIFF
--- a/android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/RibCoroutineWorkerTest.kt
+++ b/android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/RibCoroutineWorkerTest.kt
@@ -16,7 +16,10 @@
 package com.uber.rib.core
 
 import com.google.common.truth.Truth.assertThat
+import com.jakewharton.rxrelay2.BehaviorRelay
 import com.uber.autodispose.coroutinesinterop.autoDispose
+import com.uber.rib.core.WorkerBinder.mapInteractorLifecycleToWorker
+import com.uber.rib.core.lifecycle.InteractorEvent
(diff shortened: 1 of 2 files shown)


FranAguilera:
@psteiger Can you update the RIB worker demo app with a new case for this hybrid "worker"? It's located in here https://github.com/uber/RIBs/blob/main/android/demos/rib-workers/src/main/kotlin/com/uber/rib/workers/root/main/ribworkerselection/RibWorkerSelectionInteractor.kt

android/libraries/rib-base/src/test/kotlin/com/uber/rib/core/RibCoroutineWorkerTest.kt
```diff
@@ -315,3 +355,18 @@ private class TestWorker : RibCoroutineWorker {
     }
   }
 }
+
+private class WorkerAndRibCoroutineWorker(
```
FranAguilera, line 359: Can we add a comment here regarding that this pattern is not recommended?

FranAguilera (approved):

tyvsmith (approved):

PR #632, opened 2024-02-18
`ScopeProvider.coroutineScope`: Fail silently when accessing outside of RIB scope.

psteiger:
Change `ScopeProvider.coroutineScope` throwing behavior when accessing after scope is inactive.

When `RibCoroutinesConfig.shouldCoroutineScopeFailSilentlyOnLifecycleEnded` is enabled, attempting to get an instance of `coroutineScope` after the RIB gets detached will fail silently instead of throwing an error.

Accessing it before RIB is active still throws an exception.

## Motivation

Suppose you have an interactor that starts subscribing to an Rx stream asynchronously.

```kotlin
class MyInteractor {
  override fun didBecomeActive() {
    runLater {
      someObservable.autoDispose(coroutineScope).subscribe()
    }
  }
}
```

It is possible that this Rx subscription will be triggered after interactor's already inactive. In that case, subscription will be No-Op.

This code is incorrect: `Interactor` is attempting to do work after it's inactive. In other words, `runLater` does not respect the interactor lifecycle. Still, considering this is some legacy code being migrated from Rx to Coroutines, the new code would look like the following.

```kotlin
class MyInteractor {
  override fun didBecomeActive() {
    runLater {
      someObservable.autoDispose(coroutineScope).subscribe()
    }
  }
}
```

Unlike the Rx counterpart, this code will sometimes fatally throw. If `ScopeProvider.coroutineScope` is called after the scope's inactive, it will throw `LifecycleEndedException` at call site.

Calling `coroutineScope` outside of lifecycle bounds is incorrect, but in order to allow for a smoother Rx migration to Coroutines, we are changing current implementation to deliver the exception to the `CoroutineExceptionHandler` instead of throwing at call site.

If some app decides to override the default behavior of crashing (in JVM/Android) in face of attempts to obtain coroutineScope when lifecycle is not active, one can customize `RibCoroutinesConfig.exceptionHandler` at app startup:

```kotlin
RibCoroutinesConfig.exceptionHandler = CoroutineExceptionHandler { _, throwable ->
  when (throwable) {
    is CoroutineScopeLifecycleException -> log(throwable) // only log, don't re-throw.
    else -> throw throwable
  }
}
```

### DIFF
--- a/libraries/rib-coroutines/build.gradle.kts
+++ b/libraries/rib-coroutines/build.gradle.kts
@@ -13,22 +13,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 plugins {
     id("ribs.kotlin.library")
     alias(libs.plugins.maven.publish)
 }
 
 dependencies {
-
     api(libs.autodispose.coroutines)
     api(libs.kotlinx.coroutines.android)
     api(libs.kotlinx.coroutines.rx2)
 
     compileOnly(libs.android.api)
 
+    implementation(libs.autodispose.lifecycle)
+
     testImplementation(project(":libraries:rib-base"))
     testImplementation(project(":libraries:rib-test"))
+    testImplementation(project(":libraries:rib-coroutines-test"))
     testImplementation(testLibs.junit)
     testImplementation(testLibs.mockito)
     testImplementation(testLibs.mockito.kotlin)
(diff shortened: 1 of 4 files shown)


android/gradle/libs.versions.toml
```diff
@@ -33,7 +33,7 @@ javapoet = "1.11.1"
 jsr250 = "1.0"
 junit = "4.12"
 kotlin = "1.8.20"
-kotlinx-coroutines = "1.7.3"
+kotlinx-coroutines = "1.8.0-RC2"
```
sanchezz93, line 36: This is now 1.9.0-RC

jbarr21 (approved):

dmellorahul (approved):

PR #641, opened 2024-09-20
increase interactor lifecycleFlow buffer size

rysh88:
This increases the lifecycle flow buffer size in the interactor. The existing size drops emissions when there are too many subscribers. The change to the unlimited buffer will maximize the chance of successful emissions.


tyvsmith:
Please provide validation info. Any risks or concerns? New memory issues from unlimited growing list?

psteiger:
Can you please clarify which use case demanded this change?

Yes, buffer is 1 and can miss events, but did you actually go through some case where this was actually an issue?

PR #642, opened 2024-10-12
Add tests for RibEvents buffer size configuration

psteiger:


android/libraries/rib-base/src/main/kotlin/com/uber/rib/core/RibEvents.kt
```diff
@@ -20,20 +20,52 @@ import io.reactivex.Observable
...
+  private var extraBufferCapacity: Int = Channel.UNLIMITED
 
-  private val mutableRouterEvents =
-    MutableSharedFlow<RibRouterEvent>(0, Channel.UNLIMITED, BufferOverflow.DROP_OLDEST)
-  private val mutableRibDurationEvents =
-    MutableSharedFlow<RibActionInfo>(0, Channel.UNLIMITED, BufferOverflow.DROP_OLDEST)
+  /**
+   * Sets the extra buffer capacity for [routerEventsFlow] and [ribActionEventsFlow].
+   *
+   * This function must be called on the main thread, and before any usage of:
```
tyvsmith, line 33: What will happen if this isn't true? How can we guarantee it's called before?

psteiger: @tyvsmith
> What will happen if this isn't true?
Then, the original value (UNLIMITED) will be used at the moment someone calls one of the `RibEvents` flow / observable.

> How can we guarantee it's called before?
 I propose we just do it on our `Application`.

tyvsmith: What's the earliest we can safely fetch a param? Do we ever call this prior?

psteiger: @tyvsmith
> What's the earliest we can safely fetch a param?
It seems we'll just have to bite that first session with some default value, and have the new configuration for the next app starts.

> Do we ever call this prior?
No, I don't think there's any access to `RibEvents` flow before `Application.onCreate`. AFAIK only `ContentProvider` can run before, and we don't have any touching `RibEvents`.

tyvsmith (approved):

PR #651, opened 2025-07-29
Replace the SharedFlow with a StateFlow in the Interactor class

rysh88:
Using SharedFlow can lead to ANRs when the consumer is slow and the producer emits on the main thread.
Replacing it reduces the likelihood of ANRs and improves traceability when they do occur.
This will be a seamless change outside the Interactor.

### DIFF
--- a/libraries/rib-base/src/main/kotlin/com/uber/rib/core/FlowAsScope.kt
+++ b/libraries/rib-base/src/main/kotlin/com/uber/rib/core/FlowAsScope.kt
@@ -38,17 +38,17 @@ import kotlinx.coroutines.rx2.rxCompletable
  *    [range].
  */
 @CoreFriendModuleApi
-public fun <T : Comparable<T>> SharedFlow<T>.asScopeCompletable(
+public fun <T : Comparable<T>> SharedFlow<T?>.asScopeCompletable(
   range: ClosedRange<T>,
   context: CoroutineContext = EmptyCoroutineContext,
 ): CompletableSource {
   ensureAlive(range)
   return rxCompletable(RibDispatchers.Unconfined + context) {
-    takeWhile { it < range.endInclusive }.collect()
+    takeWhile { it == null || it < range.endInclusive }.collect()
   }
 }
 
-private fun <T : Comparable<T>> SharedFlow<T>.ensureAlive(range: ClosedRange<T>) {
+private fun <T : Comparable<T>> SharedFlow<T?>.ensureAlive(range: ClosedRange<T>) {
   val lastEmitted = replayCache.lastOrNull()
   when {
     lastEmitted == null || lastEmitted < range.start -> throw LifecycleNotStartedException()
--- a/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
+++ b/libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
@@ -28,8 +28,10 @@ import javax.inject.Inject
 import kotlin.properties.ReadWriteProperty
 import kotlin.reflect.KProperty
 import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.rx2.asObservable
 
 /**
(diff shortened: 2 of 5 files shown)


libraries/rib-base/src/main/kotlin/com/uber/rib/core/FlowAsScope.kt
```diff
@@ -38,17 +38,19 @@ import kotlinx.coroutines.rx2.rxCompletable
...
-public fun <T : Comparable<T>> SharedFlow<T>.asScopeCompletable(
+public fun <T : Comparable<T>> SharedFlow<T?>.asScopeCompletable(
   range: ClosedRange<T>,
   context: CoroutineContext = EmptyCoroutineContext,
 ): CompletableSource {
   ensureAlive(range)
   return rxCompletable(RibDispatchers.Unconfined + context) {
-    takeWhile { it < range.endInclusive }.collect()
+    takeWhile {
+      it == null || it < range.endInclusive
```
psteiger, line 48: Should we use something like `CREATED` for the initial state?

libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -42,9 +44,21 @@ public abstract class Interactor<P : Any, R : Router<*>>() : InteractorType, Rib
   @Inject public lateinit var injectedPresenter: P
 
   @CoreFriendModuleApi public var actualPresenter: P? = null
-  private val _lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  public open val lifecycleFlow: SharedFlow<InteractorEvent>
-    get() = _lifecycleFlow
+
+  private var useStateFlow = true
```
psteiger, line 48: Who sets this? I don't see it being set

libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
```diff
@@ -264,7 +265,7 @@ private fun getJobCoroutineContext(
 }
 
 private fun <T : Comparable<T>> Worker.bind(
-  lifecycle: SharedFlow<T>,
+  lifecycle: Flow<T>,
```
psteiger, line 268: I think leaving it as `SharedFlow` is better because shared flows never complete and that's useful information to have in this method

libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
rysh88: This will be set by each app and possibly in the app delegate

libraries/rib-base/src/main/kotlin/com/uber/rib/core/WorkerBinder.kt
rysh88: We can revert this to SharedFlow when we introduce `CREATED`

libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -42,9 +44,21 @@ public abstract class Interactor<P : Any, R : Router<*>>() : InteractorType, Rib
...
 
   @CoreFriendModuleApi public var actualPresenter: P? = null
-  private val _lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  public open val lifecycleFlow: SharedFlow<InteractorEvent>
-    get() = _lifecycleFlow
+
+  private val useStateFlow = RibEvents.useStateFlowInteractorEvent
+
+  private val _lifecycleFlow: MutableSharedFlow<InteractorEvent?> =
+    if (useStateFlow)
```
psteiger, line 51: Suggest to use brackets if multi-line if-else

libraries/rib-base/src/main/kotlin/com/uber/rib/core/Interactor.kt
```diff
@@ -42,9 +44,21 @@ public abstract class Interactor<P : Any, R : Router<*>>() : InteractorType, Rib
   @Inject public lateinit var injectedPresenter: P
 
   @CoreFriendModuleApi public var actualPresenter: P? = null
-  private val _lifecycleFlow = MutableSharedFlow<InteractorEvent>(1, 0, BufferOverflow.DROP_OLDEST)
-  public open val lifecycleFlow: SharedFlow<InteractorEvent>
-    get() = _lifecycleFlow
+
+  private val useStateFlow = RibEvents.useStateFlowInteractorEvent
```
psteiger, line 48: Hm, we can probably save an allocation here. Let's just use `get() = RibEvents.useStateFlowInteractorEvent`

psteiger (approved):

PR #665, opened 2026-09-23
Add Predictive Back support

gpolak:
**Description**:
- adds support for Predictive Back
- this is the first Iteration that adds functional support that replicates current pre-PB behavior: swiping back pops to the previous screen on the stack rather than exiting the app when PB is forced
- future changes will enhance this with support for previous screen "peek" animations

**Old RIBs + Predictive Back opt-out**
- current behavior
- works as expected
https://github.com/user-attachments/assets/6077cd8a-086a-4d16-9576-161486643fc3

**Old RIBs + NO Predictive Back opt-out**
- current Rider with old RIBs
- swiping back exits the app
https://github.com/user-attachments/assets/58621020-d78b-4042-b312-f56087b3d153

**This Change + Predictive Back opt-out**
- works as expected
- no Rider changes other than new RIB library
https://github.com/user-attachments/assets/6654d5f9-4b0d-419d-80c5-2268b5bfde9b

**This Change + NO Predictive Back opt-out**
- works as expected
- Rider has new RIBs and removes the PB opt out
https://github.com/user-attachments/assets/6d357d38-aaf9-4c07-a6d1-48395db98ce7


libraries/rib-android/src/main/kotlin/com/uber/rib/core/RibActivity.kt
```diff
@@ -50,6 +51,39 @@ public abstract class RibActivity :
   RxActivityEvents {
   private var router: ViewRouter<*, *>? = null
 
+  private val ribBackPressCallback =
```
psteiger: I'd consider this class implementing this interface to save the object allocation; not sure if that is still relevant on modern android, but it doesn't hurt

psteiger (approved):

